PR #1, opened 2018-12-12
Added implementation of the default toggleable trait

HarkonenBade:
Adds an implementation of the software toggleable trait from embedded-hal to the pins from the gpio module.

### DIFF
--- a/examples/blinky_delay.rs
+++ b/examples/blinky_delay.rs
@@ -30,10 +30,7 @@ fn main() -> ! {
         let mut delay = Delay::new(cp.SYST, clocks);
 
         loop {
-            led.set_high();
-            delay.delay_ms(1_000_u16);
-
-            led.set_low();
+            led.toggle();
             delay.delay_ms(1_000_u16);
         }
     }
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -57,7 +57,7 @@ macro_rules! gpio {
         pub mod $gpiox {
             use core::marker::PhantomData;
 
-            use embedded_hal::digital::{InputPin, OutputPin, StatefulOutputPin};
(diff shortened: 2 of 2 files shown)


HarkonenBade:
I believe that I have tested it on a Nucleo-F042K6 board. Though I can do a 100% check tomorrow to make sure that i'd definitely got the version using toggle to flash (I'd subbed toggle into the blinker example while getting things setup, so it hadn't actually made a functional difference).

therealprof:
I've modified the `blinky_delay` examples and the compiler got rid of the division emulation needed for the delay which is rather weird, shrinking the code from 624B to 274B. I'd be very happy if that was the case but it sounds to be too good to be true.

HarkonenBade:
That is rather worrying. I'll do a double check on my side tomorrow and see how things are behaving.

therealprof:
I'm travelling and don't have a STM32F042 handy. If you don't get to it first (and please also confirm the size with/without `toggle()` via `tools/capture_example_bloat.sh` to eliminate the possibility of another SNAFU) I'll check at home this weekend.

HarkonenBade:
Ok, I've tested it on my nucleo board and using toggle definitely still functions appropriately and i've confirmed that its actually hitting hardware by reflashing a few times with different delays and observing appropriate changes in flash rates. I've attached the bloat reports for with and without the changes.
[harkonenbade-master.txt](https://github.com/stm32-rs/stm32f0xx-hal/files/2680144/harkonenbade-master.txt)
[upstream.txt](https://github.com/stm32-rs/stm32f0xx-hal/files/2680145/upstream.txt)

HarkonenBade:
Ty!

PR #2, opened 2018-12-15
Add fully erased pins & Input traits for OpenDrain Output ports

david-sawatzke:
This enables pins from different Ports to be stored together. It also enables using OpenDrain pins as inputs, which is useful for multiple reasons, for example software i2c implementations. 

This code is adapted from my own stm32f030-hal crate, which also has a few other changes. It was tested there, but untested here as I don't own any 042 devices. 

Something else that caught my eye was that the 2 ```internal_pull_up``` functions have differing signatures for no discernible reason. 

I might also try a stm32f030 implementation. I'd like there to be different definitions for each different package type, as there are different pins, uarts and other peripherals available, but this seems to need a lot more code. What do you think?

### DIFF
--- a/examples/blinky_multiple.rs
+++ b/examples/blinky_multiple.rs
@@ -19,10 +19,10 @@ fn main() -> ! {
         let gpiob = p.GPIOB.split();
 
         /* (Re-)configure PA1 as output */
-        let mut led1 = gpioa.pa1.into_push_pull_output();
+        let led1 = gpioa.pa1.into_push_pull_output();
 
         /* (Re-)configure PB1 as output */
-        let mut led2 = gpiob.pb1.into_push_pull_output();
+        let led2 = gpiob.pb1.into_push_pull_output();
 
         /* Constrain clocking registers */
         let rcc = p.RCC.constrain();
--- a/examples/blinky_multiple.rs
+++ b/examples/blinky_multiple.rs
@@ -1,0 +1,52 @@
+#![no_main]
+#![no_std]
+
+use panic_halt;
+
+use stm32f0xx_hal as hal;
+
+use crate::hal::delay::Delay;
+use crate::hal::prelude::*;
+use crate::hal::stm32;
+
+use cortex_m::peripheral::Peripherals;
+use cortex_m_rt::entry;
+
+#[entry]
+fn main() -> ! {
+    if let (Some(p), Some(cp)) = (stm32::Peripherals::take(), Peripherals::take()) {
+        let gpioa = p.GPIOA.split();
+        let gpiob = p.GPIOB.split();
+
+        /* (Re-)configure PA1 as output */
+        let mut led1 = gpioa.pa1.into_push_pull_output();
+
+        /* (Re-)configure PB1 as output */
+        let mut led2 = gpiob.pb1.into_push_pull_output();
+
+        /* Constrain clocking registers */
+        let rcc = p.RCC.constrain();
+
+        /* Configure clock to 8 MHz (i.e. the default) and freeze it */
+        let clocks = rcc.cfgr.sysclk(8.mhz()).freeze();
+
+        /* Get delay provider */
+        let mut delay = Delay::new(cp.SYST, clocks);
+
+        /* Store them together */
+        let mut leds = [led1.downgrade().downgrade(), led2.downgrade().downgrade()];
+        loop {
+            leds[0].set_high();
+            leds[1].set_high();
+            delay.delay_ms(1_000_u16);
+
+            leds[0].set_low();
+            leds[1].set_low();
+            delay.delay_ms(1_000_u16);
+        }
+    }
+
+    loop {
+        continue;
+    }
+}
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -119,6 +119,17 @@ macro_rules! gpio {
                 }
             }
 
+            impl InputPin for $PXx<Output<OpenDrain>> {
+                fn is_high(&self) -> bool {
+                    !self.is_low()
+                }
+
+                fn is_low(&self) -> bool {
+                    // NOTE(unsafe) atomic read with no side effects
+                    unsafe { (*$GPIOX::ptr()).idr.read().bits() & (1 << self.i) == 0 }
+                }
+            }
+
             impl<MODE> InputPin for $PXx<Input<MODE>> {
                 fn is_high(&self) -> bool {
                     !self.is_low()
@@ -408,6 +419,17 @@ macro_rules! gpio {
                     fn set_low(&mut self) {
                         // NOTE(unsafe) atomic write to a stateless register
                         unsafe { (*$GPIOX::ptr()).bsrr.write(|w| w.bits(1 << ($i + 16))) }
+                    }
+                }
+
+                impl InputPin for $PXi<Output<OpenDrain>> {
+                    fn is_high(&self) -> bool {
+                        !self.is_low()
+                    }
+
+                    fn is_low(&self) -> bool {
+                        // NOTE(unsafe) atomic read with no side effects
+                        unsafe { (*$GPIOX::ptr()).idr.read().bits() & (1 << $i) == 0 }
                     }
                 }
 
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -49,8 +49,8 @@ pub struct Output<MODE> {
 /// Push pull output (type state)
 pub struct PushPull;
 
-use hal::digital::{InputPin, OutputPin, StatefulOutputPin};
-use stm32;
+use crate::stm32;
+use embedded_hal::digital::{InputPin, OutputPin, StatefulOutputPin};
 
 /// Fully erased pin
 // We can just pretend it's gpioa. It's modified using the bits and it can only be constructed out of already existing pins
@@ -119,7 +119,7 @@ macro_rules! gpio {
             use crate::stm32::RCC;
             use super::{
                 Alternate, Floating, GpioExt, Input, OpenDrain, Output,
-                PullDown, PullUp, PushPull, AF0, AF1, AF2, AF3, AF4, AF5, AF6, AF7,
+                PullDown, PullUp, PushPull, AF0, AF1, AF2, AF3, AF4, AF5, AF6, AF7, Pin
             };
 
             /// GPIO parts
@@ -203,7 +203,7 @@ macro_rules! gpio {
                 /// This is useful when you want to collect the pins into an array where you
                 /// need all the elements to have the same type
                 pub fn downgrade(self) -> Pin<Input<MODE>> {
-                    use stm32::gpioa;
+                    use crate::stm32::gpioa;
                     use core::intrinsics::transmute;
                     Pin {
                         i: self.i,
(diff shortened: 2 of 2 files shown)


therealprof:
@david-sawatzke 

The CI build failure is kind of interesting. I'll have a look.

> I might also try a stm32f030 implementation. I'd like there to be different definitions for each different package type, as there are different pins, uarts and other peripherals available, but this seems to need a lot more code. What do you think?

Why's that? The largest package is just a superset of the smaller ones. With very few exceptions this should just work.

david-sawatzke:
My fault. Accidentally built the wrong thing to check and it's still all 2015 edition code, so it won't compile.

therealprof:
@david-sawatzke That and you'll need to import the `Pin` type from `super` into the macro.

therealprof:
The use of `transmute` is also a bit meh...

therealprof:
If you move the degraded pin manipulation into the port block of the macro you can simply use the port itself without having to store a magic pointer in the `Pin`.

david-sawatzke:
Regarding the packages (or more specifically variations): I like that it basically guarantees that you can't use peripherals/pins that don't exist on your particular mcu, without having to verify it in the datasheet, so you can't shoot yourself in the foot. I think that's one of the major appeals of rust. 

I'm not quite sure what you mean regarding the port manipulation. I can't use something like dynamic dispatching since the ports are  completely different types, right?

therealprof:
> Regarding the packages (or more specifically variations): I like that it basically guarantees that you can't use peripherals/pins that don't exist on your particular mcu, without having to verify it in the datasheet, so you can't shoot yourself in the foot. I think that's one of the major appeals of rust.

Well, the peripherals are actually there, there's just no way to connect external hardware to it. There's no difference whether you chose an available pin but nothing is connected to it or a pin which is not available externally: you'll have to ensure that your external stuff is correctly connected either way and I do not think that the little extra error checking the addition of packages would be worth the lots of trouble this would cause.

> I'm not quite sure what you mean regarding the port manipulation. I can't use something like dynamic dispatching since the ports are completely different types, right?

If you move it inside the port specific part of the macro you can use the pointer to the port specific register in the implementation via the passed-in variable, like in the rest of the implementation. You can also check other implementations, e.g. `stm32f1xx-hal` which does it that way. Storing it inside the `Pin` is not ideal because you might be wasting memory that way, if anything that should be a `PhantomData` (which I would actually prefer since the macros are a bit unwieldy already and annoying to work with).

david-sawatzke:
> Well, the peripherals are actually there, there's just no way to connect external hardware to it.

The peripherals may be physically on the die but not according to the datasheet, and I'd rather not rely on that (especially for something like timers, which are not implemented yet). Another thing is that the AF mappings change. On the smaller packages PA2 is usart1 RX, but on lqfp64 and up it's usart2 RX, on the smaller packages i2c1 can be mapped to PA9/PA10, but not on the larger ones (again, according to the datasheet, haven't verified real behaviour).

> If you move it inside the port specific part of the macro you can use the pointer to the port specific register in the implementation via the passed-in variable, like in the rest of the implementation. You can also check other implementations, e.g. stm32f1xx-hal which does it that way

AFAICT, there is nothing like this in the stm32f1xx-hal crate (and the other crates in this organization). They only have the partially erased pins, but no case where you can get PA1 and PB3 (for example) to be the same type.

How would a generic Pin type work without storing a reference to the port in it? (Admittedly it could also be some kind of enum or port number, thus needing less space)

therealprof:
@david-sawatzke Fair enough, if it's a cluster-f* like on the STM32F030 (just checked, it really is!) then yeah, we should treat the packages differently.

> AFAICT, there is nothing like this in the stm32f1xx-hal crate (and the other crates in this organization). They only have the partially erased pins, but no case where you can get PA1 and PB3 (for example) to be the same type.

Hm, I see.

> How would a generic Pin type work without storing a reference to the port in it? (Admittedly it could also be some kind of enum or port number, thus needing less space)

How to do that with a ZST instead of a hard pointer is left as an exercise to the reader. ;)

Not sure it actually is possible but it would be nice if it was.

david-sawatzke:
I implemented the port using an enum, but it seems to me that it generates the same binary (thus using the same amount of ram) and i find the using the pointer directly to be more clear. You could also encode port and pin in the same u8 value, but the resulting obfuscation leads to more unreadable code and probably prevents some optimizations

HarkonenBade:
Hmm, the casts to `gpioa::RegisterBlock` make me feel uncomfortable. Do we have guarantees that the `RegisterBlock` will always remain identical for different ports? I'd be tempted to do the enumeration, but rather than cast it, i'd match in the trait functions and use the different ports there. Happy to re-tool the code to do this if people agree that this might be a nicer way to do it.

HarkonenBade:
Also relating to the device differences, could feature gates be used to retain the same trait implementations for things like `USART1` but alter the definitions of `Pins` for different micros?

therealprof:
@HarkonenBade 

> Hmm, the casts to gpioa::RegisterBlock make me feel uncomfortable. Do we have guarantees that the RegisterBlock will always remain identical for different ports? I'd be tempted to do the enumeration, but rather than cast it, i'd match in the trait functions and use the different ports there. Happy to re-tool the code to do this if people agree that this might be a nicer way to do it.

Yeah, I agree.

> Also relating to the device differences, could feature gates be used to retain the same trait implementations for things like USART1 but alter the definitions of Pins for different micros?

Yes, that's the main idea behind those multi-HAL-impls.

HarkonenBade:
I'm taking a look at the datasheets now, as I want to check if only the package affects the AF assignment, or if we would need to get into things like an `stm32f042Kx` feature. If its only based on package, then we might be able to get away with `stm32f042` for capabilities and `Kx` for pin layout as separate features.

david-sawatzke:
>  Do we have guarantees that the RegisterBlock will always remain identical for different ports?

It's pretty unlikely that this will change. You can see that in the LL Hal from ST they also depend on the registers having the same offsets.

```
void LL_GPIO_SetPinMode(GPIO_TypeDef *GPIOx, uint32_t Pin, uint32_t Mode)
```
(```GPIO_TypeDef``` is the struct that represents the different registers inside of a port)

I haven't seen anything in the ST32 chips where they have multiple register maps for variations of the same Peripheral, so I think that risk is neglible.

HarkonenBade:
I am very much a fan of the entirely erased pins though, as recently in some playing around I was forced to cast references to pins to `&mut OutputPin` in order to store them in an array, as I was using pins from both `gpiob` and `gpiof` due to the layout on my dev board.

HarkonenBade:
hmm, that is fair. What are the downsides to doing the trait based enumeration? Other than somewhat larger source size?

HarkonenBade:
I was aiming for something like:
```
fn is_set_low(&self) -> bool {
    match self.port {
        Gpioa => unsafe{ stm32::GPIOA::ptr().bsrr.write(|w| w.bits(1 << (self.i + 16))) },
        Gpiof => /* Set for this port*/, 
    }
}
```
Under the full assumption that currently all the `/* Set for this port*/` lines will be identical. It feels more idiomatically rust to do it this way. You could potentially even reduce the pin structure into an enumeration.

```
pub enum Pin<MODE> {
    Gpioa(u8, PhantomData<MODE>),
   /* etc*/
}
```

therealprof:
Extra kudos go to the person who manages to achieve that feat without making the macros even messier. ;)

david-sawatzke:
I strongly dislike that way, it makes the code really unreadable when using different chips with different gpio ports: 
```
fn is_set_low(&self) -> bool {
    match self.port {
        Gpioa => unsafe{ stm32::GPIOA::ptr().bsrr.write(|w| w.bits(1 << (self.i + 16))) },
        Gpiof => /* Set for this port*/, 
        #[cfg(any(feature = "stm32f030r8", feature = "stm32f030rc"))]
       Gpioc => /* Set for this port*/,
       #[cfg(any(
            feature = "stm32f030c6",
            feature = "stm32f030c8",
            feature = "stm32f030cc",
       ))]
       Gpiod => /* Set for this port*/,
    }
}
```
And that for every function. It doesn't have to be this fine grained, but you'll still need something similar because (for example) gpiod doesn't even exist on stm32f042 chips, so otherwise you'll get errors. It would also unnecessary duplicate the code at least 5 times.

therealprof:
@david-sawatzke Ideally you have one or multiple structs with custom traits implementing the HAL traits. Then you simply define the available structs for the particular chip using config gates. I agree that except for a few special cases (like clock tree setup) there shouldn't be any cfg gates inside the impls.

david-sawatzke:
> Would it be viable to have a `GPIO` trait that we implement for the various `gpioa`, `gpiob` etc, and then store in the `Pin` structure a `*const GPIO`?

I've looked at that and found no easy way, especially since the register functions like ```read()``` return a different type, so you'd have to abstract all that away for (in my view) little benefit. If the gpio structure shifts, then it's broken for that port. (In both the reference manual for the stm32f030 and f042 they define the offsets to be the same)

HarkonenBade:
I suppose my complain on the broken end is currently if the structure shifts then it will cause the crate to fail to build. If we have the pointer casting and a structure shifts and the pointer cast for it isn't removed, we will then be casting one struct to another that is now a different shape and have memory corruption bugs.

HarkonenBade:
Also I wonder if the code abstraction via a GPIO trait could be used to make life easier for the other functions, as we could boil things down to a single place where setting/reading bits is done and then all the various input and output traits for the different pin types all just use those common implementations?

HarkonenBade:
I'm imagining something like:
```
pub trait GPIO {
  fn read_in(&self, pos: u8) -> bool;
  fn read_out(&self, pos: u8) -> bool;
  fn set_high(&self, pos: u8);
  fn set_low(&self, pos: u8);
}
```

HarkonenBade:
I'm happy to have a go at a test implementation of this tomorrow once I have some free time and access to hardware for testing.

therealprof:
@HarkonenBade That could indeed be useful, however it would have to be a whole lot more involved than your suggestion since other peripherals don't use GPIO functions but dedicated peripheral functionality.

 It'd be also great if we could get rid of the macros to redefine everything on every GPIO.

HarkonenBade:
I suspect different approaches would be needed for doing other peripherals, but I'd hope this would solve the pin anonymisation. Also it feels like it is less essential for other peripherals given that you are less likely to need to say put multiple different SPI peripherals in an array together.

HarkonenBade:
Ok, I'll have a go at the trait based implementation tomorrow and them i'll present some code here. o7

HarkonenBade:
I have a nearly complete implementation at https://github.com/HarkonenBade/stm32f0xx-hal/tree/gpio-traits , though it currently does not build due to https://github.com/stm32-rs/stm32-rs/issues/115 and it's related issue https://github.com/rust-embedded/svd2rust/issues/268 . Hopefully I will be able to assist with the latter and get that into svd2rust. At which point my branch should hopefully build clean.

david-sawatzke:
Ok, seems good to me. Before needing to wait until the svd2rust issue is resolved in a release (maybe never), couldn't you implement the Gpio trait outside of the macro, for only the RegisterBlocks of gpioa and gpiof?

You should fix the example, only one downgrade function needed

HarkonenBade:
Macro based implementation done and put in as #4.

PR #3, opened 2018-12-16
[WIP] Add implementation for stm32f030

david-sawatzke:
Currently adds
- GPIO
- UART (partial)
- RCC

Currently it's hard to conditionally add some peripherals. I'd wrap then in a macro, especially for the uarts, since stm32f030xc has 6 of them.

The question is also how to do the multiple pin combinations for AF configurations. Individually adding them is a bit inconvenient. Is there any concrete reasoning behind having pin combinations and not just individual pins with usartRX/usartTX/i2cSDA/i2cSCL traits?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ rust:
 cache: cargo
 env:
 - MCU=stm32f042
-- MCU=stm32f040
-- MCU=stm32f040x6
-- MCU=stm32f040x8
-- MCU=stm32f040xc
+- MCU=stm32f030
+- MCU=stm32f030x6
+- MCU=stm32f030x8
+- MCU=stm32f030xc
 matrix:
   allow_failures:
     - rust: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,16 @@ rust:
   - stable
   - nightly
 cache: cargo
+env:
+- MCU=stm32f042
+- MCU=stm32f040
+- MCU=stm32f040x6
+- MCU=stm32f040x8
+- MCU=stm32f040xc
 matrix:
   allow_failures:
     - rust: nightly
   fast_finish: true
 script:
   - rustup target add thumbv6m-none-eabi
-  - cargo build --features=stm32f042
-  - cargo build --features=stm32f030
-  - cargo build --features=stm32f030x6
-  - cargo build --features=stm32f030x8
-  - cargo build --features=stm32f030xc
-  - cargo build
+  - cargo build --features=$MCU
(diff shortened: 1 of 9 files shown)


therealprof:
> I think the pin combinations does give some advantages, as unlike the traits per pin variant it stops you from being able to accidentally select for example USART1_TX and USART2_RX as your pin pair for a USART.

Huh? The per-pin-trait types would be of course generic over the peripheral. The stm32f4xx-hal has a nice example how to do it: https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/serial.rs

> Can you apply cfg gates to a block?

Yes, the problem is that you cannot just have a block anywhere in the code.

david-sawatzke:
> But again, I very much dislike macros to multiply variables into a huge hunk of code for a number of reasons so I'd very much prefer a trait-only implementation over the use of macros.

Ok, but I'd first wait to see what @HarkonenBade does for the gpios. 

> Also since the STM32F030 would be the first alternate choice I'd prefer if we could just add the empty target(s) to Cargo.toml, add it to the Travis CI config and add all required feature gates to the STM32F042 paths to make sure we can actually compile the stuff before adding more components.
s pretty empty
Did that. 

BTW: How would you do documentation? It's pretty nice to have an overview on docs.rs, but with no feature per default, it's going to be pretty empty on docs.rs

.travis.yml
```diff
@@ -3,12 +3,15 @@ rust:
...
-env:
-- MCU=stm32f042
 matrix:
   allow_failures:
     - rust: nightly
   fast_finish: true
 script:
   - rustup target add thumbv6m-none-eabi
-  - cargo build --features=$MCU
+  - cargo build --features=stm32f042
```
therealprof: You don't want to do that. Please keep as is and add the features to the `env` block.

therealprof:
> BTW: How would you do documentation? It's pretty nice to have an overview on docs.rs, but with no feature per default, it's going to be pretty empty over there.

Haven't thought about that to be honest, no idea.

adamgreig:
> BTW: How would you do documentation? It's pretty nice to have an overview on docs.rs, but with no feature per default, it's going to be pretty empty over there.

You can stick

```toml
[package.metadata.docs.rs]
features = ["feature1", "feature2"]
```

in your `Cargo.toml` to cause docs.rs to build with those features enabled, which is useful for this kind of thing (we do the same on stm32-rs).

therealprof:
@adamgreig Thanks. I'll try that with the stm32f4xx-hal crate.

david-sawatzke:
Thanks for reviewing. FYI, you're missing the license file

therealprof:
@david-sawatzke Yeah, License file got blocked by my .gitignore when moving the crate. Thanks for noticing.

PR #4, opened 2018-12-17
Fully erased pins based on trait implementations

HarkonenBade:
Result of discussion on #2. Adds an implementation of fully erased pins that works based on trait implementation. Currently uses a macro definition to assign the trait, this may be fixed later by some work done to svd2rust to make gpio register block definition more standard.
Based on and features code by @david-sawatzke from his PR #2.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -49,6 +49,62 @@ pub struct Output<MODE> {
 /// Push pull output (type state)
 pub struct PushPull;
 
+use crate::stm32;
+use embedded_hal::digital::{InputPin, OutputPin, StatefulOutputPin};
+
+/// Fully erased pin
+// We can just pretend it's gpioa. It's modified using the bits and it can only be constructed out of already existing pins
+pub struct Pin<MODE> {
+    i: u8,
+    port: *const stm32::gpioa::RegisterBlock,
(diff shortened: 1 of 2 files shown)


HarkonenBade:
What do you mean by already established names? I have provisionally changed it to `GpioRegExt` to match the -Ext suffix convention, but it cannot be combined into `GpioExt` as the traits are currently applied to different objects (`GpioExt` gets implemented for `GPIOA` and such, `GpioRegExt` gets implemented on `gpioa::RegisterBlock` and such).

src/gpio.rs
```diff
@@ -11,6 +11,13 @@ pub trait GpioExt {
     fn split(self) -> Self::Parts;
 }
 
+trait GpioRegExt {
+    fn in_low(&self, pos: u8) -> bool;
+    fn out_low(&self, pos: u8) -> bool;
```
therealprof: Actually I meant changing `out_low` to `is_set_low` and `in_low` to `is_low` to match up with the digital trait names.

HarkonenBade: Ah, fair. Will tweak.

HarkonenBade: Done

therealprof:
Thanks a lot for this work!

PR #6, opened 2018-12-18
Add more peripherals for the stm32f030

david-sawatzke:
- Serial is now structured like in https://github.com/stm32-rs/stm32f4xx-hal/pull/48/, so there is no giant macro repeating code
- I2C1 & I2C2 is there
- SPI1 & SPI2 is there

### TODO
- (Maybe) Port examples so they work on all variants (Opened a separate issue for that)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,8 +51,8 @@ stm32f042 = ["stm32f0/stm32f0x2"]
 stm32f030 = ["stm32f0/stm32f0x0"]
 stm32f030x4 = ["stm32f030x6"]
 stm32f030x6 = ["stm32f030"]
-stm32f030x8 = ["stm32f030", "stm32f0/stm32f0x0"]
-stm32f030xc = ["stm32f030", "stm32f0/stm32f0x0"]
+stm32f030x8 = ["stm32f030"]
+stm32f030xc = ["stm32f030"]
 
 [profile.dev]
 debug = true
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -269,8 +269,6 @@ where
 impl<USART, TXPIN, RXPIN> Serial<USART, TXPIN, RXPIN>
 where
     USART: Deref<Target = SerialRegisterBlock>,
-    TXPIN: TxPin<USART>,
-    RXPIN: RxPin<USART>,
 {
     /// Splits the UART Peripheral in a Tx and an Rx part
     /// This is required for sending/receiving
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1,3 +1,28 @@
+//! API for the integrated USART ports
+//!
+//! This only implements the usual asynchronous bidirectional 8-bit transfers, everything else is missing
+//!
+//! # Example
+//! Serial Echo
+//! ``` no_run
+//! use stm32f0xx_hal as hal;
+//!
+//! use crate::hal::stm32;
+//! use crate::hal::prelude::*;
+//! use crate::hal::serial::Serial;
+//! use nb::block;
+//!
+//! let mut p = stm32::Peripherals::take().unwrap();
+//!
+//! let mut led = gpioa.pa1.into_push_pull_pull_output();
+//! let rcc = p.RCC.constrain().cfgr.freeze();
+//! let mut timer = Timer::tim1(p.TIM1, Hertz(1), clocks);
+//! loop {
+//!     led.toggle();
+//!     block!(timer.wait()).ok();
+//! }
+//! ```
+
 use core::fmt::{Result, Write};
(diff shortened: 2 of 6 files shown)


src/serial.rs
```diff
@@ -83,63 +83,83 @@ pub struct Serial<USART, PINS> {
...
+    // USART6: (usart6, usart6en, apb2enr),
+}
+
+// It's s needed for the impls, but rustc doesn't recognize that
+#[allow(dead_code)]
+type SerialRegisterBlock = stm32::usart1::RegisterBlock;
+
+impl<USART> embedded_hal::serial::Read<u8> for Rx<USART>
+where
+    USART: Deref<Target = SerialRegisterBlock>,
```
HarkonenBade: This again feels kinda sketchy to me, as it is relying on all the USART peripherals continuing to use the same register block definition.

david-sawatzke: If it's going to break, we're getting a compile error, so in that's ok. In the worst case you can do some macro based approach where you can separately implement the different register blocks.

david-sawatzke:
Added all possible pin combinations and split up the pin traits for Serial

src/serial.rs
```diff
@@ -61,49 +61,55 @@ pub enum Error {
...
-impl Pins<stm32::USART2> for (gpioa::PA14<Alternate<AF1>>, gpioa::PA3<Alternate<AF1>>) {}
+usart_pins! {
+    USART2: (gpioa::PA2<Alternate<AF1>>, gpioa::PA3<Alternate<AF1>>),
+    USART2: (gpioa::PA14<Alternate<AF1>>, gpioa::PA15<Alternate<AF1>>),
+}
+#[cfg(feature = "stm32f030xc")]
+usart_pins! {
+    // TODO WTF look at this again, in the datasheet PB10 is both tx and rx
+    // USART3: (gpiob::PB10, AF4, gpiob::PA11, AF4),
+    USART3: (gpioc::PC4<Alternate<AF1>>, gpioc::PC5<Alternate<AF1>>),
```
therealprof: Instead of specifying pairs wouldn't it make more sense to actually single out the TX/RX functions and let the user chose any possible combination? Using this macro in this way means you're passing on a few possible combinations, e.g. you use also use PC10 and PC5 for USART3, or PC4 and PC11.

david-sawatzke: Thats the whole purpose of the separate Rx and Tx traits, the combinations are done by the templates. The only reason i defined them pairwise is because that what ST seems to have done and it seemed cleaner this way.

therealprof: @HarkonenBade That sounds like a good idea.

david-sawatzke: @HarkonenBade Agree. We could also reuse that for the spi, where the pin combinations aren't that clear.

HarkonenBade: Done, if @david-sawatzke could pull the fresh commit from https://github.com/HarkonenBade/stm32f0xx-hal/tree/stm32f030_v2 that would be great.

HarkonenBade: I took some slight liberties with making the macro syntax a a bit more verbose, but I think it makes it a lot clearer what is going on.

david-sawatzke: @HarkonenBade Yeah, that's better. The syntax also seems more "rusty".

therealprof:
Is this still WIP? Looks pretty much complete to me.

david-sawatzke:
I'd also like to port spi & i2c to the same approach as serial and add the 2nd instance

david-sawatzke:
Currently getting a rustc panic
```
thread 'main' panicked at 'byte index 1177 is out of bounds of `crate::stm32::$SPI`', libcore/str/mod.rs:2094:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

```

Cargo.toml
```diff
@@ -51,8 +51,8 @@ stm32f042 = ["stm32f0/stm32f0x2"]
 stm32f030 = ["stm32f0/stm32f0x0"]
 stm32f030x4 = ["stm32f030x6"]
 stm32f030x6 = ["stm32f030"]
-stm32f030x8 = ["stm32f030", "stm32f0/stm32f0x0"]
-stm32f030xc = ["stm32f030", "stm32f0/stm32f0x0"]
+stm32f030x8 = ["stm32f030"]
+stm32f030xc = ["stm32f030"]
```
therealprof: Snuck in another unrelated change, eh? ;)

david-sawatzke: It's technically not *completely* unrelated

PR #9, opened 2018-12-19
Add initial STM32F070 support

jessebraham:
Alright, so I did my best to just duplicate #3 as much as possible with this.

Basically just adds the minimal required support for the `STM32F070X6` and `STM32F070XB` devices. There *are* in fact hardware differences between the two families (which is why they are represented as separate features), they just have not been reflected in code yet.

If there are any changes to be made please just let me know and I'll be glad to do so :smile:


src/timers.rs
```diff
@@ -221,13 +221,17 @@ timers! {
     TIM17: (tim17, tim17en, tim17rst, apb2enr, apb2rstr),
 }
 
-#[cfg(any(feature = "stm32f030x8", feature = "stm32f030xc"))]
+#[cfg(any(
+    feature = "stm32f030x8",
+    feature = "stm32f030xc",
+    feature = "stm32f070",
+))]
```
david-sawatzke: These aren't available on stm32f070x6 devices

src/timers.rs
```diff
@@ -221,13 +221,17 @@ timers! {
...
 timers! {
     TIM6: (tim6, tim6en, tim6rst, apb1enr, apb1rstr),
     TIM15: (tim15, tim15en, tim15rst, apb2enr, apb2rstr),
 }
 
-#[cfg(feature = "stm32f030xc")]
+#[cfg(any(feature = "stm32f030xc", feature = "stm32f070"))]
```
david-sawatzke: This timer isn't available on stm32f070x6 devices as well

david-sawatzke:
You'll probably want to wait until https://github.com/stm32-rs/stm32f0xx-hal/pull/6 is merged for the serial ports.
(Probably for i2c and spi as wel)

jessebraham:
Awesome, thanks for taking the time to look it over. I can get the timer issue resolved later today, and I'll keep an eye on #6 and make the appropriate changes to Serial and take a swing at I2C/SPI once it's been merged in.

jessebraham:
Okay, first off I merged in #6 just to get my branch current. I then made the modifications to the 3 Timers @david-sawatzke  pointed out, removing support for the `STM32F070X6` devices. I then re-implemented Serial using the new methods merged in.

Please let me know if anybody spots any discrepancies. Once the above has been approved I intend to implement I2C and SPI as well; should those be in the same pull request or should we merge this first and create a separate pull request?

PR #12, opened 2018-12-20
Add a watchdog implementation

david-sawatzke:
Should be ready to merge

### DIFF
--- a/examples/watchdog.rs
+++ b/examples/watchdog.rs
@@ -1,0 +1,60 @@
+#![no_main]
+#![no_std]
+
+use panic_halt;
+
+use core::fmt::Write;
+use stm32f0xx_hal as hal;
+
+use crate::hal::delay::Delay;
+use crate::hal::prelude::*;
+use crate::hal::serial::Serial;
+use crate::hal::stm32;
+use crate::hal::time::Hertz;
+use crate::hal::watchdog::Watchdog;
+
+use cortex_m::peripheral::Peripherals;
+use cortex_m_rt::entry;
+
+#[entry]
+fn main() -> ! {
+    if let (Some(p), Some(cp)) = (stm32::Peripherals::take(), Peripherals::take()) {
+        let gpioa = p.GPIOA.split();
+        let rcc = p.RCC.constrain();
+        let dbgmcu = p.DBGMCU;
+
+        // Disable the watchdog when the cpu is stopped under debug
+        dbgmcu.apb1_fz.modify(|_, w| w.dbg_iwdg_stop().set_bit());
+
+        let mut watchdog = Watchdog::new(p.IWDG);
+        let clocks = rcc.cfgr.sysclk(8.mhz()).freeze();
+
+        // Get delay provider
+        let mut delay = Delay::new(cp.SYST, clocks);
+
+        let tx = gpioa.pa9.into_alternate_af1();
+        let rx = gpioa.pa10.into_alternate_af1();
+
+        let serial = Serial::usart1(p.USART1, (tx, rx), 115_200.bps(), clocks);
+
+        let (mut tx, _rx) = serial.split();
+        tx.write_str("RESET \r\n").ok();
+
+        watchdog.start(Hertz(1));
+        delay.delay_ms(500_u16);
+        watchdog.feed();
+        delay.delay_ms(500_u16);
+        watchdog.feed();
+        delay.delay_ms(500_u16);
+        tx.write_str("This will get printed \r\n").ok();
+        watchdog.feed();
+
+        // Now a reset happens while delaying
+        delay.delay_ms(1500_u16);
+        tx.write_str("This won't\r\n").ok();
+    }
+
+    loop {
+        continue;
+    }
+}
--- a/src/watchdog.rs
+++ b/src/watchdog.rs
@@ -114,8 +114,7 @@ impl watchdog::WatchdogEnable for Watchdog {
         self.iwdg.rlr.write(|w| w.rl().bits(time.reload));
         // Wait until the registers are updated before issuing a reset with
         // (potentially false) values
-        while self.iwdg.sr.read().pvu().bit() {}
-        while self.iwdg.sr.read().rvu().bit() {}
+        while self.iwdg.sr.read().bits() != 0 {}
         self.iwdg.kr.write(|w| w.key().reset());
     }
 }
--- a/src/watchdog.rs
+++ b/src/watchdog.rs
@@ -1,0 +1,121 @@
+//! API for the IWDG
+//!
+//! You can activate the watchdog by calling `start` or the setting appropriate
+//! device option bit when programming.
+//!
+//! After activating the watchdog, you'll have to regularly `feed` the watchdog.
+//! If more time than `timeout` has gone by since the last `feed`, your
+//! microcontroller will be reset.
+//!
+//! This is useful if you fear that your program may get stuck. In that case it
+//! won't feed the watchdog anymore, the watchdog will reset the microcontroller
+//! and thus your program will function again.
+//!
+//! **Attention**:
+//!
+//! The IWDG runs on a separate 40kHz low-accuracy clock (30kHz-60kHz). You may
+//! want to some buffer in your interval.
+//!
+//! Per default the iwdg continues to run even when you stopped execution of code via a debugger.
+//! You may want to disable the watchdog when the cpu is stopped
+//!
+//! ``` ignore
+//! let dbgmcu = p.DBGMCU;
+//! dbgmcu.apb1_fz.modify(|_, w| w.dbg_iwdg_stop().set_bit());
+//! ```
+//!
+//! # Example
+//! ``` no_run
+//! use stm32f0xx_hal as hal;
+//!
+//! use crate::hal::stm32;
+//! use crate::hal::prelude::*;
+//! use crate::hal:watchdog::Watchdog;
+//! use crate::hal:time::Hertz;
+//!
+//! let mut p = stm32::Peripherals::take().unwrap();
+//!
+//! let mut iwdg = Watchdog::new(p.iwdg);
+//! iwdg.start(Hertz(100));
+//! loop {}
+//! // Whoops, got stuck, the watchdog issues a reset after 10 ms
+//! iwdg.feed();
+//! ```
+use embedded_hal::watchdog;
+
+use crate::stm32::IWDG;
+use crate::time::Hertz;
+
+/// Watchdog instance
+pub struct Watchdog {
+    iwdg: IWDG,
+}
+
+impl watchdog::Watchdog for Watchdog {
+    /// Feed the watchdog, so that at least one `period` goes by before the next
+    /// reset
+    fn feed(&mut self) {
+        self.iwdg.kr.write(|w| w.key().reset());
+    }
+}
+
+/// Timeout configuration for the IWDG
+#[derive(PartialEq, PartialOrd, Clone, Copy)]
+pub struct IwdgTimeout {
+    psc: u8,
+    reload: u16,
+}
+
+impl Into<IwdgTimeout> for Hertz {
+    /// This converts the value so it's usable by the IWDG
+    /// Due to conversion losses, the specified frequency is a maximum
+    ///
+    /// It can also only represent values < 10000 Hertz
+    fn into(self) -> IwdgTimeout {
+        let mut time = 40_000 / 4 / self.0;
+        let mut psc = 0;
+        let mut reload = 0;
+        while psc < 7 {
+            reload = time;
+            if reload < 0x1000 {
+                break;
+            }
+            psc += 1;
+            time /= 2;
+        }
+        // As we get an integer value, reload is always below 0xFFF
+        let reload = reload as u16;
+        IwdgTimeout { psc, reload }
+    }
+}
+impl Watchdog {
+    pub fn new(iwdg: IWDG) -> Self {
+        Self { iwdg }
+    }
+}
+
+impl watchdog::WatchdogEnable for Watchdog {
+    type Time = IwdgTimeout;
+    fn start<T>(&mut self, period: T)
+    where
+        T: Into<IwdgTimeout>,
+    {
+        let time: IwdgTimeout = period.into();
+        // Feed the watchdog in case it's already running
+        // (Waiting for the registers to update takes sometime)
+        self.iwdg.kr.write(|w| w.key().reset());
+        // Enable the watchdog
+        self.iwdg.kr.write(|w| w.key().start());
+        self.iwdg.kr.write(|w| w.key().enable());
+        // Wait until it's safe to write to the registers
+        while self.iwdg.sr.read().pvu().bit() {}
+        self.iwdg.pr.write(|w| w.pr().bits(time.psc));
+        while self.iwdg.sr.read().rvu().bit() {}
+        self.iwdg.rlr.write(|w| w.rl().bits(time.reload));
+        // Wait until the registers are updated before issuing a reset with
+        // (potentially false) values
+        while self.iwdg.sr.read().pvu().bit() {}
+        while self.iwdg.sr.read().rvu().bit() {}
+        self.iwdg.kr.write(|w| w.key().reset());
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,3 +18,4 @@ pub mod serial;
 pub mod spi;
 pub mod time;
 pub mod timers;
+pub mod watchdog;
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -1,4 +1,7 @@
 pub use embedded_hal::prelude::*;
+// TODO for some reason, watchdog isn't in the embedded_hal prelude
+pub use embedded_hal::watchdog::Watchdog as _stm32f0xx_hal_embedded_hal_watchdog_Watchdog;
+pub use embedded_hal::watchdog::WatchdogEnable as _stm32f0xx_hal_embedded_hal_watchdog_WatchdogEnable;
 
 pub use crate::gpio::GpioExt as _stm32f0xx_hal_gpio_GpioExt;
 pub use crate::rcc::RccExt as _stm32f0xx_hal_rcc_RccExt;


src/watchdog.rs
```diff
@@ -0,0 +1,103 @@
...
+//! loop {}
+//! // Whoops, got stuck, the watchdog issues a reset after 10 ms
+//! iwdg.feed();
+//! ```
+use embedded_hal::watchdog;
+
+use crate::stm32::IWDG;
+use crate::time::Hertz;
+
+pub struct Watchdog {
```
HarkonenBade: Doc tags on these structs would be nice, just so they get descriptions in lists in the docs.

src/watchdog.rs
```diff
@@ -0,0 +1,103 @@
...
+impl watchdog::Watchdog for Watchdog {
+    /// Feed the watchdog, so that at least one `period` goes by before the next
+    /// reset
+    fn feed(&mut self) {
+        self.iwdg.kr.write(|w| w.key().reset());
+    }
+}
+
+#[derive(PartialEq, PartialOrd, Clone, Copy)]
+pub struct IwdgTime {
```
HarkonenBade: and second.

david-sawatzke:
@therealprof Is there any special reason why the watchdog traits may not be in the embedded_hal prelude? Otherwise I'll open an issue

HarkonenBade:
It's probably just an oversight, the ADC traits are also missing from their prelude.

david-sawatzke:
We have an unfortunate overflow in systick delay. It's handled by a debug_assert, but on real hardware you often can only run release builds. In case the internal clock is 48 MHz, the systick overflows at ~310 ms.

Watchdog seems to work fine

HarkonenBade:
> We have an unfortunate overflow in systick delay. It's handled by a debug_assert, but on real hardware you often can only run release builds. In case the internal clock is 48 MHz, the systick overflows at ~310 ms.

Should we consider capping overflowing times with an assert for during debug?

david-sawatzke:
We do, but debug builds often don't fit on target hardware

HarkonenBade:
Ah, sorry poor formatting. I meant should we cap all times to stop them overflowing, and retain the assert so you get a warning when you are in debug. But if you have built release it just gives you either what you asked for or the longest possible delay at that clockrate.

david-sawatzke:
Opened #14 for that, implementing delays too large as multiple smaller delays

therealprof:
@david-sawatzke Most likely an oversight that the prelude stuff is missing. Please open a ticket there.

therealprof:
I'll check the details of the implementation later, I've used the watchdog on the STM32F042 before so that should be easy. ;)

src/watchdog.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl watchdog::WatchdogEnable for Watchdog {
+    type Time = IwdgTimeout;
+    fn start<T>(&mut self, period: T)
+    where
+        T: Into<IwdgTimeout>,
+    {
+        let time: IwdgTimeout = period.into();
+        // First kick the watchdog
```
therealprof: Please find a different phrase than "kick the watchdog" in the comments. I know it gets used a lot (and I've certainly done it in the past) but some people find that offensive.

```
/// Feeds an existing watchdog to ensure the processor isn't reset. Sometimes
/// commonly referred to as "kicking" or "refreshing".
#[cfg(feature = "unproven")]
pub trait Watchdog {
    /// Triggers the watchdog. This must be done once the watchdog is started
    /// to prevent the processor being reset.
    fn feed(&mut self);
}
```

david-sawatzke:
The only differences I see are the additional resets at beginning and end (since I don't know when i get called, this function can also be used to reconfigure the watchdog) and the wait until it's safe to write the timeout/psc registers.

david-sawatzke:
> Most likely an oversight that the prelude stuff is missing. Please open a ticket there.

Opened https://github.com/rust-embedded/embedded-hal/issues/117

therealprof:
The RM specifies how to use the watchdog so unless there's some evidence that the manual is wrong I'd prefer if we stayed on the specified path. The example code does this:

```
/* (1) Activate IWDG (not needed if done in option bytes) */
/* (2) Enable write access to IWDG registers */
/* (3) Set prescaler by 8 */
/* (4) Set reload value to have a rollover each 100ms */
/* (5) Check if flags are reset */
/* (6) Refresh counter */
IWDG->KR = IWDG_START; /* (1) */
IWDG->KR = IWDG_WRITE_ACCESS; /* (2) */
IWDG->PR = IWDG_PR_PR_0; /* (3) */
IWDG->RLR = IWDG_RELOAD; /* (4) */
while (IWDG->SR) /* (5) */
{
    /* add time out here for a robust application */
}
IWDG->KR = IWDG_REFRESH; /* (6) */
```

therealprof:
Oh well, the example is actually wrong... the RM also says:
```
If several reload, prescaler, or window values are used by the application, it is mandatory to wait until RVU bit is reset before changing the reload value, to wait until PVU bit is reset before changing the prescaler value, and to wait until WVU bit is reset before changing the window value. However, after updating the prescaler and/or the reload/window value it is not necessary to wait until RVU or PVU or WVU is reset before continuing code execution except in case of low-power mode entry.
```

😅

But at least there's no need to feed the watchdog since a change of the reload register will do that by itself.

therealprof:
@david-sawatzke According to RM0091 this seems to be only valid when using windows mode. Must have read it somewhere else then, since this one specifically says to feed it when not operating in windowed mode... But found another bug in the RM, steps 4 and 5 of the initialisation are swapped. 🙄

david-sawatzke:
libopencm3 seems to use it the same way. http://libopencm3.org/docs/latest/stm32f0/html/group__iwdg__file.html#ga9fddf4170e5fb9d3b4df309a152e27e5
So they probably got the examples consistently wrong.

therealprof:
Hm, libopencm3 seems right from what I see but also spots some inefficiencies like setting the KR twice.

src/watchdog.rs
```diff
@@ -16,6 +16,14 @@
 //! The IWDG runs on a separate 40kHz low-accuracy clock (30kHz-60kHz). You may
 //! want to some buffer in your interval.
 //!
+//! Per default the iwdg continues to run even when you stopped execution of code via a debugger.
+//! You may want to disable the watchdog when the cpu is stopped
+//!
+//! ```
+//! let dbgmcu = p.DBGMCU;
+//! dbgmcu.apb1_fz.modify(|_, w| w.dbg_iwdg_stop().set_bit());
+//! ````
```
therealprof: Wrong use of backticks, also this might be interpreted as a standalone (even runnable) doctest which is probably not what we want.

examples/watchdog.rs
```diff
@@ -0,0 +1,60 @@
...
+        let rcc = p.RCC.constrain();
+        let dbgmcu = p.DBGMCU;
+
+        // Disable the watchdog when the cpu is stopped under debug
+        dbgmcu.apb1_fz.modify(|_, w| w.dbg_iwdg_stop().set_bit());
+
+        let mut watchdog = Watchdog::new(p.IWDG);
+        let clocks = rcc.cfgr.sysclk(8.mhz()).freeze();
+
+        /* Get delay provider */
```
therealprof:
```suggestion
        // Get delay provider
```

src/delay.rs
```diff
@@ -47,17 +47,29 @@ impl DelayMs<u8> for Delay {
 
 impl DelayUs<u32> for Delay {
```
therealprof: Changes in here don't seem watchdog related

david-sawatzke: Accidentally commited in the wrong branch (twice in a row)

david-sawatzke:
Wasn't that the issue with mismatching examples in the reference manual?

The Register descriptions say 

> The RVU bit in the IWDG_SR register must be reset in order to be able to change the reload
value.

and 
> If several reload, prescaler, or window values are used by the application, it is mandatory to
wait until RVU bit is reset before changing the reload value, to wait until PVU bit is reset
before changing the prescaler value, and to wait until WVU bit is reset before changing the
window value. However, after updating the prescaler and/or the reload/window value it is not
necessary to wait until RVU or PVU or WVU is reset before continuing code execution
except in case of low-power mode entry.

while their step by step instructions do the exact opposite 

> 4. Write the reload register (IWDG_RLR).
> 5. Wait for the registers to be updated (IWDG_SR = 0x0000 0000).

and libopencm3 does it like the register descriptions say, while stm32cube does it like in the examples.

As far as i (now) understand it, when the values are written, we have to wait quite some time. So to be safe,  we should wait before setting them and before issuing a feed. (The user might change the iwdg values somewhere else)

therealprof:
The examples (and suggested order) don't match the documentation. The documentation also explains why you can't change the values if the bits are not in reset state (i.e. the watchdog hardware is updating them): because they might be overwritten in the back and thus not take effect as intended by the programmer.

The correct way to update the settings is to check the status register **before** setting the values.

david-sawatzke:
And we have to check afterwards to make sure our registers have the right values when we feed the watchdog again.

therealprof:
@david-sawatzke Huh? Where are you getting that from?

david-sawatzke:
As far as I understand it, the delay is the time until the iwdg internal registers are updated. So when we issue a feed before that, we could potentially get the wrong refresh value from the (not yet updated) internal registers.

david-sawatzke:
Example still runs fine

src/watchdog.rs
```diff
@@ -0,0 +1,121 @@
...
+        self.iwdg.kr.write(|w| w.key().start());
+        self.iwdg.kr.write(|w| w.key().enable());
+        // Wait until it's safe to write to the registers
+        while self.iwdg.sr.read().pvu().bit() {}
+        self.iwdg.pr.write(|w| w.pr().bits(time.psc));
+        while self.iwdg.sr.read().rvu().bit() {}
+        self.iwdg.rlr.write(|w| w.rl().bits(time.reload));
+        // Wait until the registers are updated before issuing a reset with
+        // (potentially false) values
+        while self.iwdg.sr.read().pvu().bit() {}
```
therealprof: Why not wait for the whole register to be zero at once?

david-sawatzke: Seems clearer to me that we're waiting for the updates of these two registers.

therealprof: If it's a safety measure to make sure you're not feeding the watchdog with wrong values you might as well wait for the whole status to be in reset (you could actually do the same for the other waits as well, should result in less binary code).

david-sawatzke: Smaller binaries is a good reason

src/watchdog.rs
```diff
@@ -0,0 +1,119 @@
...
+        // Feed the watchdog in case it's already running
+        // (Waiting for the registers to update takes sometime)
+        self.iwdg.kr.write(|w| w.key().reset());
+        // Enable the watchdog
+        self.iwdg.kr.write(|w| w.key().start());
+        self.iwdg.kr.write(|w| w.key().enable());
+        // Wait until it's safe to write to the registers
+        while self.iwdg.sr.read().bits() != 0 {}
+        self.iwdg.pr.write(|w| w.pr().bits(time.psc));
+        self.iwdg.rlr.write(|w| w.rl().bits(time.reload));
```
therealprof:
```suggestion
        while self.iwdg.sr.read().bits() != 0 {}
        self.iwdg.rlr.write(|w| w.rl().bits(time.reload));
```

Still need to the read twice. ;)

therealprof: Because that is what the documentation says... ;)

```
If several reload, prescaler, or window values are used by the application, it is mandatory to wait until RVU bit is reset before changing the reload value, to wait until PVU bit is reset
before changing the prescaler value, and to wait until WVU bit is reset before changing the window value. However, after updating the prescaler and/or the reload/window value it is not necessary to wait until RVU or PVU or WVU is reset before continuing code execution
except in case of low-power mode entry.
```

The whole documentation is a bit ambiguous but if you want to play **really** safe you need to check the status **before** updating the values and if you're paranoid then also before resetting the reload value. I guess that none of that is actually a hard requirement...

therealprof:
I've tested the implementation with the additional wait, not the most efficient due to the additional sanity checks but works fine. If you could add the last bit and squash again I'd merge this.

david-sawatzke:
Ok, this has to be one of the most robust implementations for iwdg out there now. We're only missing timeouts.

david-sawatzke:
No, I just reverted the stuff for the registers back to how it was before. Since we only have to wait for the rvu bit before writing, it's (time-wise) wasteful if we wait on the pvu bit as well. And it's a bit confusing to have two different delay loops before setting registers

therealprof:
@david-sawatzke Ah, okay. Sorry to be of more bother but you might want to check and tweak the commit messages here.

therealprof:
> Since we only have to wait for the rvu bit before writing, it's (time-wise) wasteful if we wait on the pvu bit as well

Not really, in reality that is pretty much never of relevance anyway. It's more likely you'll loose a cycle or two for the additional instructions for extracting the bit to look at...

david-sawatzke:
(At least in the watchdog example) The pvu is set after writing to the prescaler after writing to it (as expected) so we would probably wait at least a whole 40kHz cycle unnecessarily before writing to the rlr and that would be way more than a few cycles. 

> Sorry to be of more bother but you might want to check and tweak the commit messages here.

You mean the squashed commits?

therealprof:
> (At least in the watchdog example) The pvu is set after writing to the prescaler after writing to it (as expected) so we would probably wait at least a whole 40kHz cycle unnecessarily before writing to the rlr and that would be way more than a few cycles.

I don't think that'd be the case but even if that wouldn't be a big issue as long as we manage to feed the watchdog in time. I could measure it but I'm not in the mood killing an hour or so for setting up a test.

Maybe we can promote the sanest and cleanest (and most overkill) IWDG implementation in the world... 😏

david-sawatzke:
I tested it. It is the case

therealprof:
> You mean the squashed commits?

Yes, it it lacks a bit of sleekness. ;)

```
Add an example

Fix documentation

Fix race condition

Improve wording of watchdog introduction

Rename the watchdog time struct so it's clearer

Add doc comments for structs

Tolerate clock innacuracies

Use a lower clock value, otherwise SYSTICK will overflow

Use feed when referring to the watchdog

Add mention of iwdg under debug

Fix code sample for iwdg under debug

Use consistent comments
```

david-sawatzke:
Out of some reason only the 64 bit variant gets inlined in blinky_multiple (both  without special cases: 
With `(us * self.clocks.sysclk().0 / 1_000_000`:
``` 
~/u/p/181127-embedded_rust-69/stm32f0xx-hal [delay] λ cargo bloat --features stm32f042 --example blinky_multiple --release
File  .text Size         Crate Name
0.0%   0.0%   0B               [0 Others]
0.2%  33.7% 288B     [Unknown] main
0.1%  16.9% 144B           std __udivmodsi4
0.1%  14.8% 126B           std __clzsi2
0.0%   8.4%  72B stm32f0xx_hal <stm32f0xx_hal::delay::Delay as embedded_hal::blocking::delay::DelayMs<u16>>::delay_ms
0.0%   7.5%  64B   cortex_m_rt Reset
0.0%   2.3%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_set_low
0.0%   2.3%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_low
0.0%   2.3%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_set_low
0.0%   2.3%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_low
0.0%   2.1%  18B     [Unknown] __aeabi_uidivmod
0.0%   1.6%  14B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_low
0.0%   1.6%  14B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_low
0.0%   1.4%  12B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_high
0.0%   1.4%  12B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_high
0.0%   0.2%   2B   cortex_m_rt HardFault_
0.0%   0.2%   2B           std core::ptr::drop_in_place
0.0%   0.2%   2B           std core::ptr::drop_in_place
0.0%   0.2%   2B   cortex_m_rt DefaultPreInit
0.0%   0.2%   2B   cortex_m_rt DefaultHandler_
0.5% 100.0% 854B               .text section size, the file size is 179.6KiB
```
With `            ((us as u64 * self.clocks.sysclk().0 as u64) / 1_000_000) as u32`
```
~/u/p/181127-embedded_rust-69/stm32f0xx-hal [delay] λ cargo bloat --features stm32f042 --example blinky_multiple --release
Compiling ...
Analyzing target/thumbv6m-none-eabi/release/examples/blinky_multiple

File  .text Size         Crate Name
0.0%   0.0%   0B               [0 Others]
0.3%  52.0% 288B     [Unknown] main
0.1%  11.6%  64B   cortex_m_rt Reset
0.1%  10.8%  60B stm32f0xx_hal <stm32f0xx_hal::delay::Delay as embedded_hal::blocking::delay::DelayMs<u16>>::delay_ms
0.0%   3.6%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_set_low
0.0%   3.6%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_low
0.0%   3.6%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_set_low
0.0%   3.6%  20B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::is_low
0.0%   2.5%  14B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_low
0.0%   2.5%  14B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_low
0.0%   2.2%  12B stm32f0xx_hal <stm32f0::stm32f0x2::gpiof::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_high
0.0%   2.2%  12B stm32f0xx_hal <stm32f0::stm32f0x2::gpioa::RegisterBlock as stm32f0xx_hal::gpio::GpioRegExt>::set_high
0.0%   0.4%   2B   cortex_m_rt HardFault_
0.0%   0.4%   2B           std core::ptr::drop_in_place
0.0%   0.4%   2B           std core::ptr::drop_in_place
0.0%   0.4%   2B   cortex_m_rt DefaultPreInit
0.0%   0.4%   2B   cortex_m_rt DefaultHandler_
0.5% 100.0% 554B               .text section size, the file size is 111.1KiB
```

Issue #11, opened 2018-12-19
Implement ADC functionality


HarkonenBade:
Add functionality for reading ADC values.

HarkonenBade:
I'm planning to make a go at this at some point soon, though no idea how successful i'll be.

therealprof:
ADC is very useful to read analog sensors (or the operating voltage of the MCU). I tend to prefer digital sensors where possible or binarised signals using hardware but for newcomers from the Arduino world this is a must have.

HarkonenBade:
Got some first tentative steps https://github.com/HarkonenBade/stm32f0xx-hal/tree/adc-test, gonna throw them at some hardware tomorrow and see if it even remotely works.

therealprof:
NB: I do have the hardware to generate accurate signals and measure them.

HarkonenBade:
Noted, I think I can also get hold of some of that, though for now my hopes are around 'any reading' possibly climbing as high as 'a small number when V = 0v and a large number when V = 3v3'

HarkonenBade:
Gonna work on actually implementing all the channels. Might also add some bits for getting the vref and temp if i'm feeling especially fancy.

therealprof:
> Might also add some bits for getting the vref and temp if i'm feeling especially fancy.

Those are actually quite useful because they make decent out-of-the-box examples.

PR #13, opened 2018-12-20
Initial implementation of ADC interface

HarkonenBade:
Adds an initial implementation of oneshot ADC conversion.


HarkonenBade:
Closes #11.

src/adc.rs
```diff
@@ -0,0 +1,268 @@
...
+impl VTemp {
+    pub fn new() -> Self { VTemp {} }
+
+    /// Enable the internal temperature sense, this has a wake up time
+    /// t<sub>START</sub> which can be found in your micro's datasheet, you
+    /// must wait at least that long after enabling before taking a reading.
+    /// Remember to disable when not in use.
+    pub fn enable(&mut self) {
+        let adc = unsafe{ &*stm32::ADC::ptr() };
+        adc.ccr.write(|w| w.tsen().set_bit());
```
david-sawatzke: Why prohibit leaving multiple on at a time? (of course, then you'll have to consider race conditions)

src/adc.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+impl<WORD, PIN> OneShot<Adc, WORD, PIN> for Adc
+    where
+        WORD: From<u16>,
+        PIN: Channel<Adc, ID=u8>,
+{
+    type Error = ();
+
+    fn read(&mut self, _pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
```
david-sawatzke: I'd at least add a comment that not all bits are used. 
I'd prefer just shifting it and leaving the lower bits always at zero, seems like a less surprising interface to me.

src/adc.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /// Disable the internal reference voltage.
+    pub fn disable(&mut self) {
+        let adc = unsafe{ &*stm32::ADC::ptr() };
+        adc.ccr.write(|w| w.vrefen().clear_bit());
+    }
+}
+
+impl VBat {
```
david-sawatzke: VBat doesn't exist on stm32f030, should probably be behind a featuregate

src/adc.rs
```diff
@@ -0,0 +1,268 @@
...
+        self.rb.cr.write(|w| w.adcal().set_bit());
+
+        /* Wait until calibration is finished and ADCAL = 0 */
+        while self.rb.cr.read().adcal().bit_is_set() {}
+    }
+
+    fn select_clock(&mut self) {
+        let rcc = unsafe { &*stm32::RCC::ptr()};
+
+        rcc.apb2enr.write(|w| w.adcen().set_bit());
```
david-sawatzke: Shouldn't this be a modify?

src/adc.rs
HarkonenBade: The default setup does this.

src/adc.rs
HarkonenBade: Ah, noted. Will gate.

src/adc.rs
HarkonenBade: Ah, I have misunderstood how to operate the registers slightly, will correct. Most of my uses of `write` should instead by `modify`.

src/adc.rs
HarkonenBade: As it is setup by default the 12 bits get stored in the lower twelve bits of the data register, essentially producing a u16 with a value of 0-4095. The lower precision modes also shift right as appropriate so they also present as 0-n values.

david-sawatzke: No, i mean exactly the opposite. Few people like reading docs and per default I'd expect the u16 value to go from 0 to 0xFFFF (technically 0xFFF0 in this case). This also makes the trait non-portable, since  potential libraries would probably expect a normal u16 value.

HarkonenBade: Ah, so we would lose the least 4 significant bits. That is definitely do-able, though introduces a weird quirk at 6bit res.

HarkonenBade: Possibly I could include an optional mode if you want instead numbers always from 0-N. I was planning anyway on adding something that displayed the max range value.

examples/blinky_adc.rs
```diff
@@ -0,0 +1,57 @@
...
+        let mut adc = Adc::new(p.ADC);
+
+        loop {
+            led.toggle();
+
+            let val: u16 = adc.read(&mut an_in).unwrap();
+
+            /* shift the value right by 3, same as divide by 8, reduces
+               the 0-4095 range into something approximating 1-512 */
+            let time: u16 = (val >> 3) + 1;
```
david-sawatzke: The threshhold has changed (in the doc test as well)

therealprof:
Hm, something seems off. I've tried to read the temperature sensor but I'm getting absurd values (i.e. way outside the calibration values).

HarkonenBade:
Hmm, did you ensure that you waited for t_start time before prodding the temperature sensor? Also are the stored calibration values left or right aligned?

therealprof:
> Hmm, did you ensure that you waited for t_start time before prodding the temperature sensor?

I haven't checked the t_start time but I would assume that a few seconds should be enough.
The calibration values are 16 bit values and I've checked that they are read okay and make sense (by plugging those into the formulas to arrive at 110°C and 30°C respectively). They are 1325 (for 110°C) and 1792 (for 30°) while the ADC readings end up somewhere around 27920, so they're off by a factor of 16ish...

therealprof:
NB: You're missing channel 18 for V_bat.

david-sawatzke:
> Factor of 16ish

They are probably right aligned, while our values are left alinged, so that would make sense as that make a factor of 16 with 12 bits

david-sawatzke:
Our abstraction is completely useless for non mcu-specific code otherwise. No code can guess that we only use the lowest few bits while returning a u16 (not the whole range).

HarkonenBade:
The awkward bit is that all the MCU provided calibration bits are going to be right aligned then. Also @therealprof VBat should be available on stm32f04x and stm3207x devices, but I had gated it out on stm3203x devices as @david-sawatzke mentioned that it wasnt present there. Also the discussion on alignment is in one of the earlier revolved conversaions.

david-sawatzke:
Vbat also doesn't exist on stm32070 (it's the same RM)

therealprof:
> Our abstraction is completely useless for non mcu-specific code otherwise. No code can guess that we only use the lowest few bits while returning a u16 (not the whole range).

I disagree. The difference between left-aligned and right-aligned is qualitative vs. quantitative measurements. If you're just interested in (resolution independent) fractions the whole range then you go left aligned, if you're interested in the actual value then you go right-aligned.

Not that I would care much but Arduino does right-aligned by default so you will find a lot of users which would expect the same here.

In general the alignment and also the resolution should be a users choice and not fixed by the implementation. 

> Also @therealprof VBat should be available on stm32f04x and stm3207x devices, but I had gated it out on stm3203x

No worries, that can be added later.

HarkonenBade:
@therealprof @david-sawatzke I think this is in just about final shape. I'd appreciate some glances over and input. The only extensions I have currently in mind are adding functions to read things like the TSense value as degrees C using the RM example code, however I am uncertain as to how good/bad using floating point code will be, so I might have to take a careful look at that. Also not sure if offering a builder interface for configuring an Adc at creation would be appreciated or not.

HarkonenBade:
Also I'm happy to do one last pass and flatten it down to one commit before merge.

therealprof:
@HarkonenBade Looking great! Please add to CHANGELOG.md, `cargo fmt` it and squash into fewer commits.

HarkonenBade:
Roger

HarkonenBade:
@therealprof Done, though there are a few tiny issues remaining. It doesn't currently build clean with no features, and I'm not quite sure how to wrangle that. Also the cargo fmt run wanted me to shift a huge pile of feature statements in other files from being split over lines into being stacked on one line, I have ignored those but it does mean it doesn't fmt clean.

PR #17, opened 2018-12-22
Stm32f04x additions for RCC

bentwire:
Please review.  There are some leftovers from debugging still around, I will remove them.

I'm not quite sure how to do the feature flags for the clock builder, I could really use some help there, not sure how to set the code up.  It is why I added flags to the RCC struct that I have not used yet...

Thanks!

Mike


bentwire:
Oh I also have some code I modified from https://github.com/ah-/anne-key to work with the USB on the stm32f072, but I don't know if this is the correct crate for it, or if it should go in to a "usb-hal" crate with the code I have been messing with here: https://github.com/bentwire/stm32f072-usb

src/rcc.rs
```diff
@@ -37,12 +47,36 @@ pub struct Rcc {
 
 #[allow(unused)]
 const HSI: u32 = 8_000_000; // Hz
+#[allow(dead_code)] // Only used for ADC, but no configuration for ADC clock yet.
+const HSI14: u32 = 14_000_000; // Hz - ADC clock.
+// Can't do this anymore... #[cfg(any(feature = "stm32f042", feature = "stm32f072"))]
+const HSI48: u32 = 48_000_000; // Hz - (available on STM32F04x, STM32F07x and STM32F09x devices only)
+
+pub enum SysClkSource {
```
david-sawatzke: Why is this pub?

src/rcc.rs
```diff
@@ -37,12 +47,36 @@ pub struct Rcc {
...
+    HSI = 0b00,
+    HSE = 0b01,
+    PLL = 0b10,
+    HSI48 = 0b11,
+}
+
+pub enum PllSource {
```
david-sawatzke: This as well

Cargo.toml
```diff
@@ -46,6 +46,7 @@ panic-halt = "0.2.0"
 [features]
 rt = ["stm32f0/rt"]
 stm32f042 = ["stm32f0/stm32f0x2"]
+stm32f072 = ["stm32f0/stm32f0x2"]
```
david-sawatzke: I'd add stm32f072 compatability in a seperate PR. Even then, you're missing it in travis

src/rcc.rs
```diff
@@ -37,12 +47,36 @@ pub struct Rcc {
 
 #[allow(unused)]
 const HSI: u32 = 8_000_000; // Hz
+#[allow(dead_code)] // Only used for ADC, but no configuration for ADC clock yet.
+const HSI14: u32 = 14_000_000; // Hz - ADC clock.
```
david-sawatzke: If we don't need it yet (and probably not in the future, since we do configuration of HSI14 in the ADC PR), i wouldn't add it

src/rcc.rs
```diff
@@ -37,12 +47,36 @@ pub struct Rcc {
...
+    HSI48 = 0b11,
+}
 
 #[allow(unused)]
 pub struct CFGR {
     hclk: Option<u32>,
     pclk: Option<u32>,
     sysclk: Option<u32>,
+    enable_hsi: Option<bool>,
+    enable_hsi14: Option<bool>,
```
david-sawatzke: We're don't need this as well

src/rcc.rs
```diff
@@ -75,20 +109,77 @@ impl CFGR {
         self
     }
 
+    pub fn enable_hsi(mut self, is_enabled: bool) -> Self {
+        self.enable_hsi = Some(is_enabled);
+        self
+    }
+
+    pub fn enable_hsi14(mut self, is_enabled: bool) -> Self {
```
david-sawatzke: We do configuration inside the adc pr (at least in the pr)

src/rcc.rs
```diff
@@ -133,25 +224,79 @@ impl CFGR {
...
+        // HSI
+        if self.enable_hsi.unwrap_or(false) {
+            rcc.cr.write(|w| w.hsion().set_bit());
+            while rcc.cr.read().hsirdy().bit_is_clear() { }
+        }
+        // HSI14
+        if self.enable_hsi14.unwrap_or(false) {
+            rcc.cr2.modify(|_, w| w.hsi14on().set_bit());
+            while rcc.cr2.read().hsi14rdy().bit_is_clear() {
+                // nothing
```
david-sawatzke: These empty loops aren't consistent. I'd do it without the comment inside

src/rcc.rs
```diff
@@ -75,20 +109,77 @@ impl CFGR {
...
+        let sysclk = self.sysclk.unwrap_or(HSI);
 
-        let pllmul_bits = if pllmul == 2 {
-            None
+        let r_sysclk; // The "real" sysclock value, calculated below
+        let src_clk_freq; // Frequency of source clock for PLL and etc, HSI, or HSI48 on supported systems.
+        let pllmul_bits;
+
+        // Select clock source based on user input and capability
+        // Highest selected frequency source available takes precident.
```
david-sawatzke: *precedent

src/rcc.rs
```diff
@@ -75,20 +109,77 @@ impl CFGR {
...
+        // For F04x, F07x, F09x parts, use HSI48 if requested.
+        if self.enable_hsi48.unwrap_or(false) { 
+            // This does not work... #[cfg(any(feature = "stm32f042", feature = "stm32f072"))]
+            src_clk_freq = HSI48; // Use HSI48 if requested and available.
+        } else if self.enable_hsi.unwrap_or(true) {
+            src_clk_freq = HSI; // HSI if requested, or by default.
+        } else {
+            src_clk_freq = HSI; // If no clock source is selclected use HSI.
```
david-sawatzke: *selected

src/rcc.rs
```diff
@@ -37,12 +47,36 @@ pub struct Rcc {
...
 
 #[allow(unused)]
 pub struct CFGR {
     hclk: Option<u32>,
     pclk: Option<u32>,
     sysclk: Option<u32>,
+    enable_hsi: Option<bool>,
+    enable_hsi14: Option<bool>,
+    // This does not work very well: #[cfg(any(feature = "stm32f042", feature = "stm32f072"))]
+    enable_hsi48: Option<bool>,
```
david-sawatzke: Is there a reason for these (except `enable_hsi`) being an Option?

src/rcc.rs
```diff
@@ -37,12 +47,36 @@ pub struct Rcc {
...
+    HSE = 0b10,
+    HSI48 = 0b11,
+}
 
 #[allow(unused)]
 pub struct CFGR {
     hclk: Option<u32>,
     pclk: Option<u32>,
     sysclk: Option<u32>,
+    enable_hsi: Option<bool>,
```
HarkonenBade: What is the logic for making all of these booleans options? Given that it looks like later you `unwrap_or` them all later with default values. Why not just make them plain booleans and initialising them with the default values?

src/rcc.rs
bentwire: Hmm, sorry that should not be.  Oversight on my part.

src/rcc.rs
bentwire: Same here. I will remove pub from both of these.

src/rcc.rs
bentwire: Ok I will remove this then.

src/rcc.rs
bentwire: No problem I can remove it.

src/rcc.rs
bentwire: Ok cool.

src/rcc.rs
bentwire: Good point.  I was trying to be consistent, but maybe that isn't called for here.

src/rcc.rs
bentwire: Ok cool I can remove this.  I guess I just assumed all the RCC stuff would be handled in one place.   Clocks configured before peripherals, etc.

src/rcc.rs
```diff
@@ -75,20 +109,77 @@ impl CFGR {
...
+            let pllmul = (4 * self.sysclk.unwrap_or(src_clk_freq) + src_clk_freq) / src_clk_freq / 2;
+            let pllmul = cmp::min(cmp::max(pllmul, 2), 16);
+            r_sysclk = pllmul * src_clk_freq / 2;
+
+            pllmul_bits = if pllmul == 2 {
+                None
+            } else {
+                Some(pllmul as u8 - 2)
+            };
+            enable_pll = true; // Force PLL on.
```
bentwire: I think I am going to remove this as well.  enable_pll isn't needed because pllmul_bits acts as the gate for enabling the PLL...

Suggestions?

Cargo.toml
bentwire: Should I remove this then?

src/rcc.rs
david-sawatzke: Yeah, I'd rename pllmul_bits to pllmul and as that's already Option it's alright

david-sawatzke:
Yeah, adding 072 support should be a separate PR. Having this as a feature while providing no functionality is quite misleading

src/rcc.rs
bentwire: After re-reading your question and the one below I think I have this fixed now?

david-sawatzke: Yeah, looks fine now

bentwire: Awesome :)

therealprof:
Hi @bentwire. Thanks a lot for your contribution. Additional clocking support is highly welcome, especially since this is the most tricky part of MCU programming. From your other comment I gather that you would like to enable USB support which would be fantastic feat! Is there any other use case you had in mind for it?

bentwire:
> Hi @bentwire. Thanks a lot for your contribution. Additional clocking support is highly welcome, especially since this is the most tricky part of MCU programming. From your other comment I gather that you would like to enable USB support which would be fantastic feat! Is there any other use case you had in mind for it?

Yeah I am attempting to get USB working, that is my main goal, basically I want to move from C to Rust for my STM32 based board designs.  It is a hit and miss thing at the moment, ST's example code is all over the place, and the rust code I am working from is for a chip that is "just different enough" that it won't work as is :/

Thankfully though I have a Beagle 12 and a Nucleo-F072RB to test with.

Has anyone come up with a usb-hal yet?

therealprof:
> Yeah I am attempting to get USB working, that is my main goal, basically I want to move from C to Rust for my STM32 based board designs.

Fantastic. I started from the same spot just a year ago. USB is still on my bucket list, I made a secret santa project with could hugely benefit from USB on the STM32F042...

> It is a hit and miss thing at the moment, ST's example code is all over the place, and the rust code I am working from is for a chip that is "just different enough" that it won't work as is :/

Well, as usual that depends where you're coming from; if you're a heavy HAL/LL user and know everything inside out then Rust might still look a bit pedestrian. If you're coming an RTOS which only has partial support then Rust could even have a very charming appeal.

> Has anyone come up with a usb-hal yet?

There's something for the STM32F103: https://github.com/mvirkkunen/stm32f103xx-usb . From what I hear USB seems quite similar for a lot of STM32 chips and a lot of the descriptor stuff is independent of the hardware anyway so it might be possible to crate a common driver.

therealprof:
> Then I started coding in it, it is quite nice. :) I'm still learning quite a bit about the language though so please excuse some of my noobish mistakes.

No worries. We all have to start somewhere. I'm also still learning **a lot**. ;)

> I will have to take a look at that. 

I believe People are already in the process of supporting other families, too. Most of us are hanging our on IRC in #rust-embedded and discussing things like USB on STM32.

bentwire:
Was this ok to merge or did you want me to make more changes?

src/rcc.rs
```diff
@@ -37,18 +42,38 @@ pub struct Rcc {
...
+enum SysClkSource {
+    HSI = 0b00,
+    HSE = 0b01,
+    PLL = 0b10,
+    HSI48 = 0b11,
+}
+
+#[allow(unused)]
+enum PllSource {
+    HSI_2 = 0b00,
```
therealprof: Please rename this to `HSI_DIV2` to clarify what it means

therealprof:
It's adding 16 bytes of binary bloat to all applications but other than that looks great.

PR #22, opened 2018-12-27
Initial test implementation of ADC helper functions

HarkonenBade:
Test implementation of the helper functions for various ADC internal sensors. Currently untested on actual hardware.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ version = "0.2.2"
 
 [dev-dependencies]
 panic-halt = "0.2.0"
+cortex-m-semihosting = "0.3.1"
 
 [features]
 device-selected = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,6 @@ version = "0.2.2"
 
 [dev-dependencies]
 panic-halt = "0.2.0"
-cortex-m-semihosting = "0.3.1"
 
 [features]
 device-selected = []
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -43,6 +43,9 @@ const VTEMPCAL30: *const u16 = 0x1FFF_F7B8 as *const u16;
 const VTEMPCAL110: *const u16 = 0x1FFF_F7C2 as *const u16;
 
 #[cfg(feature = "device-selected")]
+const VDD_CALIB: u16 = 3300;
+
+#[cfg(feature = "device-selected")]
 use core::ptr;
 
 #[cfg(feature = "device-selected")]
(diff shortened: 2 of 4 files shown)


src/adc.rs
```diff
@@ -46,9 +58,10 @@ pub struct Adc {
     sample_time: AdcSampleTime,
     align: AdcAlign,
     precision: AdcPrecision,
+    stored_cfg: Option<(AdcSampleTime, AdcAlign, AdcPrecision)>,
```
david-sawatzke: Couldn't you just store this on the stack for the functions, e.g. the stash (should then be renamed) just gives the tuple back and restore just takes it. There's no real need to always allocate space for this (i think)

HarkonenBade: This could be done, I suspect the storage use is incredibly minimal but I can correct.

src/adc.rs
```diff
@@ -241,6 +254,52 @@ impl VTemp {
...
+            vtemp.enable(adc);
+
+            // Double read of vdda to allow sufficient startup time for the temp sensor
+            adc.read_vdda();
+        }
+        let vdda = adc.read_vdda();
+
+        adc.stash_cfg();
+
+        adc.set_align(AdcAlign::Right);
```
david-sawatzke: Do you really need to duplicate this code? Couldn't you just put this in a non-public `get_raw` method (or something)?

src/adc.rs
```diff
@@ -241,6 +254,52 @@ impl VTemp {
...
+    }
+
+    pub fn read_vtemp(adc: &mut Adc) -> i16 {
+        let mut vtemp = Self::new();
+        let vtemp_preenable = vtemp.is_enabled(&adc);
+
+        if !vtemp_preenable {
+            vtemp.enable(adc);
+
+            // Double read of vdda to allow sufficient startup time for the temp sensor
```
david-sawatzke: Wouldn't using a delay reference be nicer here?

src/adc.rs
HarkonenBade: I could put in a reset to default settings function.

src/adc.rs
HarkonenBade: Yes, though I was trying to avoid forcing the user to provide a delay if possible.

src/adc.rs
HarkonenBade: It would be saving 3 bytes on a structure that can only ever be instantiated once.

HarkonenBade: Done

src/adc.rs
HarkonenBade: Made an option

src/adc.rs
HarkonenBade: Done

HarkonenBade:
Are there any more structural features that people think should change? Otherwise I will start working on documenting and stuff to make this into a finished PR for merging.

HarkonenBade:
@therealprof @david-sawatzke Ok, should be in final shape now. Would appreciate any final comments or requests, otherwise should be good to merge.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ version = "0.2.2"
 
 [dev-dependencies]
 panic-halt = "0.2.0"
+cortex-m-semihosting = "0.3.1"
```
therealprof:
```suggestion
cortex-m-semihosting = "0.3.1"
```

Seems superfluous, no?

HarkonenBade: Apologise, it was from my previous semihosting based example

HarkonenBade:
This is what I get for trying to be fast and commit code directly on ghub :/

PR #23, opened 2018-12-27
Allow multi delay instances for the systick timer

### LINKED ISSUE #21: Potential fault in delay.rs
I was poking clippy to see if it had any useful insights and it threw up this:
```
warning: digits grouped inconsistently by underscores
  --> src/delay.rs:85:19
   |
85 |             us / (1_000_00 / self.clocks.sysclk().0)
   |                   ^^^^^^^^ help: consider: `100_000`
   |
   = note: #[warn(clippy::inconsistent_digit_grouping)] on by default
   = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/master/index.html#inconsistent_digit_grouping
```

Is this constant supposed to be 100,000 or should it but 1,000,000?

david-sawatzke:
A lot of device crates need a delay instances, see https://github.com/JohnDoneth/hd44780-driver and https://github.com/yuri91/ili9341-rs. Having to pass a delay instance as needed isn't very ergonomic. I think multiple delay instances are often needed. There are multiple ways to do this (that I can think of):

- Allow a delay instance based on asm delay loops (probably needs ramfunc because of wait states)
- Allow normal timers to be used for delays
- Allow multiple instances using the systick timer read only 

I used the last way, leaving the systick always running, getting the timer value at the beginning and comparing it to the current one (including some amount of leeway so that long interrupts can't easily overflow). The first commit also fixes #21 (just a typo). We technically don't need an instance after init,  but if the user were to reconfigure the timer, the delay would break.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Added the option to have multiple Delay instances by cloning it - @david-sawatzke
+
 ## [v0.10.1] - 2018-12-25
 
 ### Added
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -82,7 +82,7 @@ impl DelayUs<u32> for Delay {
         const MAX_RVR: u32 = 0x00FF_FFFF;
 
         let mut total_rvr = if self.clocks.sysclk().0 < 1_000_000 {
-            us / (1_000_00 / self.clocks.sysclk().0)
+            us / (1_000_000 / self.clocks.sysclk().0)
         } else {
             us * (self.clocks.sysclk().0 / 1_000_000)
         };
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -33,22 +33,22 @@ use crate::rcc::Clocks;
 use embedded_hal::blocking::delay::{DelayMs, DelayUs};
 
 /// System timer (SysTick) as a delay provider
+#[derive(Clone)]
 pub struct Delay {
     clocks: Clocks,
-    syst: SYST,
 }
 
 impl Delay {
     /// Configures the system timer (SysTick) as a delay provider
-    pub fn new(mut syst: SYST, clocks: Clocks) -> Self {
+    /// As access to the count register is possible without a reference, we can
+    /// just drop it
+    pub fn new(mut syst: SYST, clocks: Clocks) -> Delay {
         syst.set_clock_source(SystClkSource::Core);
 
-        Delay { syst, clocks }
-    }
-
-    /// Releases the system timer (SysTick) resource
-    pub fn free(self) -> SYST {
-        self.syst
+        syst.set_reload(0x00FF_FFFF);
+        syst.clear_current();
+        syst.enable_counter();
+        Delay { clocks }
     }
 }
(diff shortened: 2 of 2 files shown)


david-sawatzke:
With the latest commit c14aad1fe7c1, `led_hal_button_irq` is a bit smaller than it currently is in master, but unfortunately it doesn't even get properly inlined in `blinky_delay` anymore, compared to 6cd9711.

I've also tested it and `blinky_delay` seems to work fine

HarkonenBade:
Out of interest, would it be possible to make it so that the frozen clock state is a static singleton? So that anyone anywhere could get a delay instance independently, assuming that the clocks are configured?

therealprof:
@HarkonenBade In theory that should be possible but I think it would be best if that was part of a generic framework.

HarkonenBade:
@therealprof What do you mean exactly? As for instance it would be very handy in the ADC code to be able to potentially grab a delay instance.

therealprof:
@HarkonenBade It's not a good design for a HAL to have code block for a longer time, especially not wait for a `delay`. This is something the application should take care of and the best way to do it would be to split the crucial parts off and have the timing-sensitive part return a result so the application can try again later whenever convenient. Of course someone might spin a wrapper around that to do it in one go using a `Delay` instance.

Providing shareable timing facilities is a step in the direction of an operating facilities and those should best be done in a generic framework a lot of HALs can use. This might require additional `embedded-hal` traits...

HarkonenBade:
Yeah, after poking it seems that the embedded-hal traits specifically make delays blocking only and timers potentially non-blocking.

david-sawatzke:
I think you could simply use timers for that. They are non-blocking, but afaict there's no easy way to use them without having to mutably borrow a timer instance

therealprof:
> Ah fair, one thing I was tempted by was attempting a yielding version of the ADC code, but it would require nightly generator features or a huge amount of state tracking sadly.

If there's a way to check the condition was met, you can simply create a function which returns a `nb::Error::WouldBlock` if it's too early to retrieve the result. The user can then decide whether to busy-poll, delay and try again or do something completely different and get back later. Async is the way to go in the future but that'll be later in 2019...

The big question is: Does a flag exist which allows us to determine whether the result is stable.

HarkonenBade:
> > Ah fair, one thing I was tempted by was attempting a yielding version of the ADC code, but it would require nightly generator features or a huge amount of state tracking sadly.
> 
> If there's a way to check the condition was met, you can simply create a function which returns a `nb::Error::WouldBlock` if it's too early to retrieve the result. The user can then decide whether to busy-poll, delay and try again or do something completely different and get back later. Async is the way to go in the future but that'll be later in 2019...
> 
> The big question is: Does a flag exist which allows us to determine whether the result is stable.

For almost everything, yes. For temp sensor startup, no.

therealprof:
Yes, can't really use a timer for that either.

therealprof:
> For almost everything, yes. For temp sensor startup, no.

That's what I thought. :(

HarkonenBade:
I could be very dirty and fake it off with non-blocking ADC reads, but that feels suboptimal.

HarkonenBade:
We should also probably take this thread of discussion over to #22 or a new issue.

PR #25, opened 2018-12-29
RCC: Add ability to use HSE clock source

zklapow:
This adds a `clock_src` param to the `CFGR` builder and the accompanying code to enable the `HSE` clock in `freeze`. This supports using the HSE both by itself and as input to the PLL and can be used by simply adding a `.hse` call like so:

```rust
        let clocks = rcc.cfgr.hse(16.mhz()).sysclk(32.mhz()).freeze();
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Support for STM32F091 - @jessebraham
+- Support for HSE as a system clocksource (#25 - breaking change) - @zklapow
 
 ## [v0.11.1] - 2019-01-05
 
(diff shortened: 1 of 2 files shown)


therealprof:
@zklapow Thanks a lot for your contribution. This looks like a very nice addition. I'm not sure what the CI build issue is but I'm actually getting the same error locally after applying your PR. Also I think this will clash quite heavily with #17 so I'd like to ask you to rework your PR when this has landed.

david-sawatzke:
The issue is the empty `repro_hardfault.rs` file, where the `#[no_std]` is missing, so rust tries to compile it with std.

HarkonenBade:
> The issue is the empty `repro_hardfault.rs` file, where the `#[no_std]` is missing, so rust tries to compile it with std.

Nice catch!

zklapow:
sorry about that, did not intend to commit that file should be gone now.

@therealprof I saw that and figured as much 👍 but figured it was worth submitting this early anyhow. On that note I am working on some other things for this hal (pwm support is the big one) should I create issues for those things or just submit PRs when ready?

therealprof:
@zklapow Ping?

zklapow:
just pushed an update that should work with the new changes

zklapow:
👍 done

PR #31, opened 2019-01-04
Added definitions for the stm32f072xx chips

Yatekii:
All the definitions for the stm32f072xx chips should be added.
Even the less popular ones like the BGA 100 should be correct.
If someone has too much spare time they are invited to spend the same hours reading the datasheets and my added definitions to double check :)

I also fixed some gpio definitions for other classes of the M0 series.

Whilst doing all this I recognized that at least the ADC definitions for the 070xx line are wrong. As this and also the GPIOs are wrong I would strongly recommend to double check the definitions of all peripherals!
Also, considering the different featuresets, the feature flags for this line do not make sense.

What's more is that I saw that some pins have shared AF1 (SPI, I2S) as they share a clock iirc. This makes sense from a HW perspective but I am afraid the HAL as it is cannot handle it.

Also is it right that the UART cannot handle HW FC atm? Will this be added in the future? I don't need this feature, I just wondered.

Best,
Yatekii

Sources:
[0] Datasheet: https://www.st.com/resource/en/datasheet/stm32f072rb.pdf
[1] Family Overview: https://www.st.com/content/st_com/en/products/microcontrollers/stm32-32-bit-arm-cortex-mcus/stm32-mainstream-mcus/stm32f0-series/stm32f0x0-value-line.html?querycriteria=productId=LN1826
[2] Reference Manual (for the clock enables): https://www.st.com/content/ccc/resource/technical/document/reference_manual/c2/f8/8a/f2/18/e6/43/96/DM00031936.pdf/files/DM00031936.pdf/jcr:content/translations/en.DM00031936.pdf


Yatekii:
Did as requested.

therealprof:
LGTM, thanks.

PR #34, opened 2019-01-05
Add a tx only and a rx only serial instance

david-sawatzke:
This also makes the serial stuff in https://github.com/stm32-rs/stm32f0xx-hal/pull/27 obsolete. 
Additionally by virtue of how it works also implements https://github.com/stm32-rs/stm32f4xx-hal/pull/49 here.

Not sure how we should handle the release method. It currently works but you get a `()` for the unused pin back.

Currently untested

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -239,7 +239,9 @@ macro_rules! usart {
                 pub fn $usart(usart: $USART, pins: (TXPIN, RXPIN), baud_rate: Bps, rcc: &mut Rcc) -> Self
                 {
                     let mut serial = Serial { usart, pins };
-                    serial.enable(baud_rate, rcc);
+                    serial.configure(baud_rate, rcc);
+                    // Enable transmission and receiving
+                    serial.usart.cr1.modify(|_, w| w.te().set_bit().re().set_bit().ue().set_bit());
                     serial
                 }
             }
@@ -253,7 +255,9 @@ macro_rules! usart {
                 {
                     let rxpin = ();
                     let mut serial = Serial { usart, pins: (txpin, rxpin) };
-                    serial.enable(baud_rate, rcc);
+                    serial.configure(baud_rate, rcc);
+                    // Enable transmission
+                    serial.usart.cr1.modify(|_, w| w.te().set_bit().ue().set_bit());
                     serial
                 }
             }
@@ -267,13 +271,15 @@ macro_rules! usart {
                 {
                     let txpin = ();
                     let mut serial = Serial { usart, pins: (txpin, rxpin) };
-                    serial.enable(baud_rate, rcc);
+                    serial.configure(baud_rate, rcc);
+                    // Enable receiving
+                    serial.usart.cr1.modify(|_, w| w.re().set_bit().ue().set_bit());
                     serial
                 }
             }
 
             impl<TXPIN, RXPIN> Serial<$USART, TXPIN, RXPIN> {
-                fn enable(&mut self, baud_rate: Bps, rcc: &mut Rcc) {
+                fn configure(&mut self, baud_rate: Bps, rcc: &mut Rcc) {
                     // Enable clock for USART
                     rcc.regs.$apbenr.modify(|_, w| w.$usartXen().set_bit());
 
@@ -284,9 +290,6 @@ macro_rules! usart {
                     // Reset other registers to disable advanced USART features
                     self.usart.cr2.reset();
                     self.usart.cr3.reset();
-
-                    // Enable transmission and receiving
-                    self.usart.cr1.modify(|_, w| w.te().set_bit().re().set_bit().ue().set_bit());
                 }
 
                 /// Starts listening for an interrupt event
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -227,31 +227,66 @@ pub struct Tx<USART> {
 unsafe impl<USART> Send for Tx<USART> {}
 
 macro_rules! usart {
-    ($($USART:ident: ($usart:ident, $usartXen:ident, $apbenr:ident),)+) => {
+    ($($USART:ident: ($usart:ident, $usarttx:ident, $usartrx:ident, $usartXen:ident, $apbenr:ident),)+) => {
         $(
             use crate::stm32::$USART;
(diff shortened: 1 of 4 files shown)


HarkonenBade:
I have some thoughts on this, will put together an actual review in a bit, kinda busy atm

src/serial.rs
```diff
@@ -186,16 +186,53 @@ unsafe impl<USART> Send for Tx<USART> {}
...
+            where
+                TXPIN: TxPin<$USART>,
+            {
+                /// Creates a new tx-only serial instance
+                pub fn $usarttx(usart: $USART, txpin: TXPIN, baud_rate: Bps, clocks: Clocks) -> Self
+                {
+                    let rxpin = ();
+                    let mut serial = Serial { usart, pins: (txpin, rxpin) };
+                    serial.enable(baud_rate, clocks);
+                    serial
```
HarkonenBade: Why not have the Tx only constructor just return a `Tx<USART>`?

src/serial.rs
```diff
@@ -186,16 +186,53 @@ unsafe impl<USART> Send for Tx<USART> {}
...
+            where
+                RXPIN: RxPin<$USART>,
+            {
+                /// Creates a new tx-only serial instance
+                pub fn $usartrx(usart: $USART, rxpin: RXPIN, baud_rate: Bps, clocks: Clocks) -> Self
+                {
+                    let txpin = ();
+                    let mut serial = Serial { usart, pins: (txpin, rxpin) };
+                    serial.enable(baud_rate, clocks);
+                    serial
```
HarkonenBade: As above but for Rx

src/serial.rs
```diff
@@ -204,16 +241,14 @@ macro_rules! usart {
...
-                    usart.cr2.reset();
-                    usart.cr3.reset();
+                    self.usart.cr2.reset();
+                    self.usart.cr3.reset();
 
                     // Enable transmission and receiving
-                    usart.cr1.modify(|_, w| w.te().set_bit().re().set_bit().ue().set_bit());
-
-                    Serial { usart, pins }
+                    self.usart.cr1.modify(|_, w| w.te().set_bit().re().set_bit().ue().set_bit());
```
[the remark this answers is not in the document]
david-sawatzke: I don't think it matters, but changed it

src/serial.rs
david-sawatzke: I want to enable `listen` and potential future things for Rx/Tx only usarts. Returning a `Tx` would also make `release` pretty hard to do.

HarkonenBade: I don't see how that fixes split usarts from being unable to release their pins?

david-sawatzke: That was in reference to the second part. The first part is also party addressed by just not needing to split, but to release the pins of the split usarts would be a bit more involved. We'd need to change `Tx` & `Rx` to contain pin references and also add an api to create a pin instance from a function (which is a bit out of scope for this pr).

HarkonenBade: I suppose I tend to see splitting usarts as a useful and important feature as it allows them to be partially shared into interrupt handlers in a sensible way without lock contention.

HarkonenBade: That sounds reasonable to me, also I'm unsure over the use of generating temporary Tx/Rx instances to impl the Read/Write traits, it would seem more appropriate to move the implementations out to free functions that take a *const USART argument.

david-sawatzke: Moved it out in separate functions with RegisterBlock, I could remove the USART parameter from `Tx` & `Rx` type, not sure if I should.

HarkonenBade: I'd leave it there, its potentially relevant to distinguish Tx and Rx objects from different USART types.

src/serial.rs
```diff
@@ -420,3 +465,52 @@ where
...
+    let isr = unsafe { (*usart).isr.read() };
+
+    if isr.tc().bit_is_set() {
+        Ok(())
+    } else {
+        Err(nb::Error::WouldBlock)
+    }
+}
+
+/// Tries to write a byte to the uart
```
therealprof:
```suggestion
/// Tries to write a byte to the UART
```

src/serial.rs
```diff
@@ -420,3 +465,52 @@ where
...
+        // NOTE(unsafe) atomic write to stateless register
+        // NOTE(write_volatile) 8-bit write that's not possible through the svd2rust API
+        unsafe { ptr::write_volatile(&(*usart).tdr as *const _ as *mut _, byte) }
+        Ok(())
+    } else {
+        Err(nb::Error::WouldBlock)
+    }
+}
+
+/// Tries to read a byte from the uart
```
therealprof:
```suggestion
/// Tries to read a byte from the UART
```

therealprof:
> Actually, would it also be worth updating at least one of the examples that doesn't really need split under this? Maybe the plain echo?

I haven't looked in detail but is this even a possibility, i.e. does the implementation allow for multiple independent uses of the chosen `USART` to attach multiple independent pins to it? If so that might be construed by some as unsafeness... 😏

Anyway, I'd be okay with modifying the examples or adding another, like a plain hello world, but I strongly favour better documentation over examples.

HarkonenBade:
@therealprof It doesn't let you define multiple usarts with the same pins, but the way the impls of read and write for usarts are worded also implements them for duplex usarts. So a full duplex usart can have both read and write called on it, and thus doesn't require splits if you don't need to send the parts to different places.

therealprof:
@HarkonenBade Sorry, I misunderstood you then. Sure, go ahead.

HarkonenBade:
Though this has raised an interesting question for me, could we shift it so that `split()` returns `(Serial<USART1, TxPin, ()>, Serial<USART1, (), RxPin>)` and just eliminate `Tx` and `Rx` all together?

david-sawatzke:
I think this would make a destructor pretty hard. It also has issues with the listen/unlisten calls (the classic race condition thing). I also don't like two serials of the same port existing, for me it looks like a hack (because it is).

HarkonenBade:
That's fair, it was just a thought based on the pleasing way the read/write impls turned out.

PR #35, opened 2019-01-05
Add support for the STM32f091

jessebraham:
Adds device support for the [STM32F091](https://www.st.com/en/microcontrollers/stm32f0x1.html?querycriteria=productId=LN7). Let me know if anything needs fixing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,9 @@ stm32f070 = ["stm32f0/stm32f0x0", "device-selected"]
 stm32f070x6 = ["stm32f070"]
 stm32f070xb = ["stm32f070"]
 stm32f072 = ["stm32f0/stm32f0x2", "device-selected"]
+stm32f091 = ["stm32f0/stm32f0x1", "device-selected"]
+stm32f091xb = ["stm32f091"]
+stm32f091xc = ["stm32f091"]
 
 [profile.dev]
 debug = true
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -56,6 +56,9 @@ stm32f070 = ["stm32f0/stm32f0x0", "device-selected"]
 stm32f070x6 = ["stm32f070"]
 stm32f070xb = ["stm32f070"]
 stm32f072 = ["stm32f0/stm32f0x2", "device-selected"]
+stm32f091 = ["stm32f0/stm32f0x1", "device-selected"]
+stm32f091xb = ["stm32f091"]
+stm32f091xc = ["stm32f091"]
```
HarkonenBade: These two extra features seem to be superfluous given they don't appear in any feature gates.

jessebraham: I agree, I was just thinking that as well funnily enough. I'll remove the variants.

therealprof: And I was about to comment the same time @HarkonenBade did. 👍 

Other than that looks really great to me.

src/serial.rs
```diff
@@ -156,6 +167,19 @@ usart_pins! {
...
+// #[cfg(feature = "stm32f091")]
+// usart_pins! {
+//     USART5 => {
+//         tx => [gpioe::PE10<Alternate<AF1>>],
+//         rx => [gpioe::PE11<Alternate<AF1>>],
+//     }
+//     USART6 => {
+//         tx => [gpiof::PF9<Alternate<AF1>>],
+//         rx => [gpiof::PF10<Alternate<AF1>>],
+//     }
```
HarkonenBade: Why is this one commented out given it is GPIOF?

jessebraham: Good eye, looks like I accidentally commented out the entire macro call rather than just the `USART5` block. I'm on it.

PR #38, opened 2019-01-07
Add fixes to RCC code for plain builds

HarkonenBade:


david-sawatzke:
The plain build should probably be added to ci. Otherwise this will happen again

HarkonenBade:
I have a bit of travis changes ready for that, but need this in first to stop that bit just insta failing CI.

PR #39, opened 2019-01-07
Tweaks to travis to run plain builds and faster cargo executions

HarkonenBade:


HarkonenBade:
Cuts builds down to 5min and runs plain builds. Should still happily error out if the build fails for whatever reason.

david-sawatzke:
Probably a good idea

PR #42, opened 2019-01-09
Optimize delay implementation

david-sawatzke:
I tried to optimize the delay implementation a bit. It gets now inlined (without division) in blinky_delay and makes led_hal_button_irq a bit smaller. Unfortunately the division is still in the delay method (here for blinky_multiple):
```

0800027c <_ZN97_$LT$stm32f0xx_hal..delay..Delay$u20$as$u20$embedded_hal..blocking..delay..DelayMs$LT$u16$GT$$GT$8delay_ms17hee9179c2bd627785E>:
 800027c:	b570      	push	{r4, r5, r6, lr}
 800027e:	2801      	cmp	r0, #1
 8000280:	d105      	bne.n	800028e <_ZN97_$LT$stm32f0xx_hal..delay..Delay$u20$as$u20$embedded_hal..blocking..delay..DelayMs$LT$u16$GT$$GT$8delay_ms17hee9179c2bd627785E+0x12>
 8000282:	2900      	cmp	r1, #0
 8000284:	d018      	beq.n	80002b8 <_ZN97_$LT$stm32f0xx_hal..delay..Delay$u20$as$u20$embedded_hal..blocking..delay..DelayMs$LT$u16$GT$$GT$8delay_ms17hee9179c2bd627785E+0x3c>
 8000286:	480e      	ldr	r0, [pc, #56]	; (80002c0 <_ZN97_$LT$stm32f0xx_hal..delay..Delay$u20$as$u20$embedded_hal..blocking..delay..DelayMs$LT$u16$GT$$GT$8delay_ms17hee9179c2bd627785E+0x44>)
 8000288:	f000 f854 	bl	8000334 <__aeabi_uidiv>
```

This is caused by the optimizer leaving the `Div` case in, for some reason.

Untested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support for STM32F091 - @jessebraham
 - Support for HSE as a system clocksource (#25 - breaking change) - @zklapow
 
+### Changed
+
+- Optimize delay implemenation (#42) - @david-sawatzke
+
 ### Fixed
 
 - Fixed panic in delay overflow handling for debug builds - @david-sawatzke
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -1,7 +1,7 @@
 //! API for delays with the systick timer
 //!
-//! Please be aware of potential overflows.
-//! For example, the maximum delay with 48MHz is around 89 seconds
+//! Please be aware of potential overflows when using `delay_us`.
+//! E.g. at 48MHz the maximum delay is 89 seconds.
 //!
 //! Consider using the timers api as a more flexible interface
 //!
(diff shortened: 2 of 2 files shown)


david-sawatzke:
We could maybe add a `Delay<Mult>` and a `Delay<Div>`, so it's enforced by the type system.

Or just panic when the clock is too low, a simpler and less functional version

therealprof:
@david-sawatzke The code could use a few comments explaining what we do and why. It seems to me you're trying to cater for both very small and very large delays regardless of the clock rate. Without calculator it is not totally obvious what the ranges are.

Now the factor 1_000_000 is not that far away from the clock speeds used so we if we desperately want to avoid the division we could also do a subtraction loop to calculate the factor/divisor. Or we could change from decimal to power of 2 calculations.

Upon closer look the whole calculation is wrong anyway: `SysTick` is running on `HCLK` or `HCLK/8` not `SysClk` so we're potentially using the wrong value.

I'm pretty sure a lot of calculations will go downhill if someone seriously decides to run the system at less than 1MHz HCLK. So maybe we should just prevent that possibility altogether.

david-sawatzke:
> I'm pretty sure a lot of calculations will go downhill if someone seriously decides to run the system at less than 1MHz HCLK. So maybe we should just prevent that possibility altogether.

I think except delay & rcc, we have no issues in serial, spi, timers (and i2c). But when considering the typical "just run faster and sleep more", it's probably not a bad idea. 

> Upon closer look the whole calculation is wrong anyway: SysTick is running on HCLK or HCLK/8 not SysClk so we're potentially using the wrong value.

Fixed that.

PR #43, opened 2019-01-11
Add support for stm32f031, stm32f051, stm32f071

jessebraham:
Alright, so hopefully this is all correct :smile: 

This adds support for the above devices, and rounds out the remainder of the `stm32f0x0` line (leaving only the `stm32f0x8` line remaining!). As usual if there are any updates needed just let me know and I'll be glad to make them.

On a side note, I unfortunately don't have any of these devices on hand to test with, so I hope that's okay.


therealprof:
Very nice! I actually have a STM32F031 somewhere, at some point I ordered all that where available in 20pin TSSOP. No idea where that might be right now though. 😅

We'll figure out soon enough if it works, I guess.

HarkonenBade:
This definitely makes me want to get #29 moving though.

PR #55, opened 2019-02-05
Use modify instead of write for HSI14 clock setup

therealprof:


therealprof:
@HarkonenBade As far as I can see the HSI14 configuration is trampling over the HSI48 configuration, so if the latter is set-up first (and potentially used as SysClock) then things might get ugly. Can you confirm?

PR #62, opened 2019-06-05
Enable CRS support for USB

therealprof:

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -14,6 +14,18 @@ impl RccExt for RCC {
             pclk: None,
             sysclk: None,
             clock_src: SysClkSource::HSI,
+            /// CRS is only available on devices with USB and HSI48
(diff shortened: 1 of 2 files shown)


src/rcc.rs
```diff
@@ -322,10 +352,36 @@ impl CFGR {
...
             let sw_var = self::inner::get_sww(&self.clock_src);
 
+            // CRS is only available on devices with USB and HSI48
+            #[cfg(any(
+                feature = "stm32f042",
+                feature = "stm32f048",
+                feature = "stm32f072",
+                feature = "stm32f078",
+            ))]
+            match self.crs {
```
Disasm: `if let Some(crs) = self.crs {` ?

therealprof: Yields a nice error message that `#[cfg]` gating is not yet supported on `if`. 😅

PR #63, opened 2019-06-05
Clear error flags before returning

### LINKED ISSUE #61: Flags are never cleared when there's a UART error
If there's a UART error (for example, it receives a byte when there's already a byte in the Receive Data Register), a sticky flag gets set. This flag causes `read()` to return an error, but the flag doesn't get cleared. So every subsequent `read()` also returns an error.

I propose you clear the flag by writing to the appropriate bit in `icf` just before the error is returned.

therealprof:
Closes #61

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


thejpster:
This PR LGTM - it's pretty much what I did with my private fork.

PR #69, opened 2019-09-23
Added an example for i2c

Maldus512:
I added an example to find all addresses of connected i2c devices.
It's a little arbitrary in which it does not show any outside effect; one is meant to run it step by step and inspect the final values with gdb.

### DIFF
--- a/examples/i2c_find_address.rs
+++ b/examples/i2c_find_address.rs
@@ -33,7 +33,11 @@ fn main() -> ! {
             let mut i2c = I2c::i2c1(p.I2C1, (scl, sda), 100.khz(), &mut rcc);
 
             let mut _devices = 0;
-            for add in 0..127 {
+            // I2C addresses are 7-bit wide, covering the 0-127 range
+            for add in 0..=127 {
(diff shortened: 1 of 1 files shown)


examples/i2c_find_address.rs
```diff
@@ -0,0 +1,52 @@
...
+
+            // Configure pins for I2C
+            let sda = gpiob.pb7.into_alternate_af1(cs);
+            let scl = gpiob.pb8.into_alternate_af1(cs);
+
+            // Configure I2C with 100kHz rate
+            let mut i2c = I2c::i2c1(p.I2C1, (scl, sda), 100.khz(), &mut rcc);
+
+            let mut _devices = 0;
+            for add in 0..127 {
```
therealprof: I believe this should be 
```suggestion
            for add in 0..=127 {
```

therealprof:
I do have a bit more comprehensive I2C scanner here:
https://github.com/stm32-rs/nucleo-f042k6/blob/master/examples/i2c_scanner.rs

But I'd be happy to take this one if there were a few more comments middle section about how this works.

PR #70, opened 2019-09-23
Add dac

bentwire:
Stole the Dac code from the STM32G0xx HAL crate and modified it to work with the STM32F072.  Only tested with 072 so only the 072 feature has the driver right now.


bentwire:
Looks like I need to fix feature gating...

bentwire:
Fixed!

src/dac.rs
```diff
@@ -0,0 +1,114 @@
...
+pub struct C2;
+
+pub trait DacOut<V> {
+    fn set_value(&mut self, val: V);
+    fn get_value(&mut self) -> V;
+}
+
+pub trait DacPin {
+    fn enable(&mut self);
+//    fn calibrate(&mut self, delay: &mut Delay);
```
therealprof:
```suggestion
```

bentwire: oops!  I will get that.

bentwire:
Ok I will fix this ASAP, probably this weekend. Thanks!

bentwire:
Er forgot to hit the comment button on the last comment...  So the feature gates and CHANGELOG entry are added.  Just need to add the example.

bentwire:
Oh and docs.

therealprof:
Oh, and I just noticed it doesn't compile, most likely because because you need to feature gate the example for stm32f072, you can check out `Cargo.toml` on how to do that.

bentwire:
Hopefully this last push should fix those issues. :)  Just lost my job, have more time to work on this now!

Issue #77, opened 2019-12-01
Problem when using external clock source


katyo:
Usually I clock my precision devices from external 8MHz crystal and it fails to work here.
There is my setup:
```rust
            // Setup clock from 8 MHz external crystal and 48 MHz PLL
            let mut rcc = p.RCC
                .configure()
                .hse(8.mhz())
                .sysclk(48.mhz())
                .freeze(&mut p.FLASH);
```
There infinite busy-loop occurs on PLL ready flag awaiting.
To exclude hardware problems I tried to load code from other project in C with similar setup and it works as expected.
It seems PLL multiplier calculation is wrong because in case of using HSE as a clock source according to datasheet the input frequency does not divided by two alike a case when HSI is used.
I may mistake but that my debugger says:
```
// before enable_pll()
(gdb) svd RCC CR
Fields in RCC CR:
        HSION:     1  Internal High Speed clock enable
        HSIRDY:    1  Internal High Speed clock ready flag
        HSITRIM:  16  Internal High Speed clock trimming
        HSICAL:   93  Internal High Speed clock Calibration
        HSEON:     1  External High Speed clock enable
        HSERDY:    1  External High Speed clock ready flag
        HSEBYP:    0  External High Speed clock Bypass
        CSSON:     1  Clock Security System enable
        PLLON:     0  PLL enable
        PLLRDY:    0  PLL clock ready flag
(gdb) s
stm32f0xx_hal::rcc::inner::enable_pll (rcc=<optimized out>, c_src=0x20000fe0, ppre_bits=<optimized out>,
    hpre_bits=<optimized out>, pllmul_bits=<optimized out>)
// after enable_pll()
(gdb) svd RCC CR
Fields in RCC CR:
        HSION:     1  Internal High Speed clock enable
        HSIRDY:    1  Internal High Speed clock ready flag
        HSITRIM:  16  Internal High Speed clock trimming
        HSICAL:   93  Internal High Speed clock Calibration
        HSEON:     0  External High Speed clock enable
        HSERDY:    0  External High Speed clock ready flag
        HSEBYP:    0  External High Speed clock Bypass
        CSSON:     0  Clock Security System enable
        PLLON:     1  PLL enable
        PLLRDY:    0  PLL clock ready flag
(gdb) svd RCC CFGR
Fields in RCC CFGR:
        SW:         0  System clock Switch
        SWS:        0  System Clock Switch Status
        HPRE:       0  AHB prescaler
[...]

PR #78, opened 2019-12-02
Fixed PLL multiplier calculation when clocking from HSE

katyo:
Fix for issue #77

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -292,12 +292,21 @@ impl CFGR {
             pllmul_bits = None;
             r_sysclk = src_clk_freq;
         } else {
-            let pllmul =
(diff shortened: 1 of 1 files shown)


pigrew:
The logic for the "pll source" frequency needs to be modified to not always divide by two. It looks like only the F030 devices predivide HSI by two. In addition, the HSI48 source needs to be accounted for in the devices where it exists.

katyo:
@pigrew Of course, it is little bit more tricky work around that case.

therealprof:
I think this was also addressed by the more comprehensive #87?

matoushybl:
Can we close this PR, @therealprof @katyo?

therealprof:
I think this is superceded. Let's close this then.

PR #80, opened 2019-12-26
Add (breaking) API change to allow for HSE bypass to be selected.

### LINKED ISSUE #79: FEATURE: Allow HSE bypass clock source for F070
An ability to set a "bypass" clock source should be added. Currently, HSE always is in the non-bypassed mode when using this HAL.

I started to code this (it's fairly simple), but don't know what the desired Rcc API would be.

My first choice would be to modify the existing fn to be `pub fn hse<F>(mut self, freq: F, bypassed: bool)`. The disadvantage is that this breaks existing code.

Another implementation would be to create `pub fn hse_bypass<F>(mut self, bypassed: bool)`, with a default to use a XTAL (non-bypassed) source.

This feature is needed for most of the Nucleo boards, where the clock source is an external oscillator (the STLink). I can attempt to implement either API (though I'm very new to Rust).

pigrew:
Closes #79 

This is an API change to allow "bypass" mode to be used for the HSE.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- Added "bypass" parameter to Rcc HSE configuration (breaking change)
+
 ### Fixed
 - RCC: Correct code to enable PLL.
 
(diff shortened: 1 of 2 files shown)


pigrew:
@therealprof

Could you please review this patch? It perhaps should be ported to other STM32 HALs, too, but I'm not sure what sort of coordination is expected.

pigrew:
> Looks good to me but would you please run the code through rustfmt?

@therealprof: Thanks for the review. I've run rustfmt and CI is now rerunning.

PR #81, opened 2019-12-27
systick returns, so should be marked as such.

pigrew:
systick() should return `()`. This fixes build errors that CI has been reporting.

### DIFF
--- a/examples/adc_values.rs
+++ b/examples/adc_values.rs
@@ -70,7 +70,7 @@ fn main() -> ! {
 }
 
 #[exception]
-fn SysTick() -> ! {
+fn SysTick() {
     use core::ops::DerefMut;
 
     // Enter critical section
--- a/examples/flash_systick.rs
+++ b/examples/flash_systick.rs
@@ -57,7 +57,7 @@ fn main() -> ! {
 // Define an exception handler, i.e. function to call when exception occurs. Here, if our SysTick
 // timer generates an exception the following handler will be called
 #[exception]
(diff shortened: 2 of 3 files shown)


therealprof:
Actually that is technically not quite correct. An exception/interrupt handler does not return in the sense of a function return, the compiler error is really curious especially since I am not able to reproduce it locally.

pigrew:
My understanding was that `!` implied that execution flow would never reach the end of the function (i.e., the function contains an infinite loop), but this exception handler does finish... (though returning from an exception on ARM doesn't follow the normal function call ABI).

I'm very new to Rust, so feel to suggest a different solution? I'm running the "stable" branch of Rust on a Windows computer, for what it's worth...

therealprof:
I think this will need some investigation. I don't have a problem changing it per se but it is kind of weird (and unexpected) that it broke in the first place.

pigrew:
One reference I used was the Rust Embedded Book, where [the Exceptions ](https://doc.rust-lang.org/1.34.2/embedded-book/start/exceptions.html#exceptions) section shows `SysTick` as returning void:

```
#[exception]
fn SysTick() {
    static mut COUNT: u32 = 0;

    // `COUNT` has type `&mut u32` and it's safe to use
    *COUNT += 1;
}
```

therealprof:
This might be a side-effect of https://github.com/rust-embedded/cortex-m-rt/pull/205. I guess we should merge it then.

PR #82, opened 2019-12-27
RCC: PLL: Modify CR instead of write, plus remove unnecessary unsafe

pigrew:
The PLL was not locking on a F070RB. Part of the reason was that the CR register was written instead of modified. This, plus fixed computation of the PLL multiplier (similar to #78) and #80, allows the PLL to lock.

(If I should first create an issue for this, or should combine all small changes together into one PR, let me know).

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -87,9 +87,9 @@ mod inner {
 
         // Set PLL source and multiplier
         rcc.cfgr
-            .modify(|_, w| unsafe { w.pllsrc().bit(pllsrc_bit).pllmul().bits(pllmul_bits) });
+            .modify(|_, w| w.pllsrc().bit(pllsrc_bit).pllmul().bits(pllmul_bits));
 
-        rcc.cr.write(|w| w.pllon().set_bit());
+        rcc.cr.modify(|_, w| w.pllon().set_bit());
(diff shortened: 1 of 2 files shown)


src/rcc.rs
```diff
@@ -176,7 +176,7 @@ mod inner {
         rcc.cfgr
             .modify(|_, w| w.pllsrc().bits(pllsrc_bit).pllmul().bits(pllmul_bits));
 
-        rcc.cr.write(|w| w.pllon().set_bit());
+        rcc.cr.modify(_, |w| w.pllon().set_bit());
```
therealprof: This is obviously an incorrect change.

pigrew:
Apologies regarding my typo. In fact, I was testing the other code path (F070RB nucleo board), and didn't immediately notice the issue (confounded by the `SysTick()` errors I'm getting).

With this patch, #79, and a fix for #77 (which caused a factor of 2 error in the PLL multiplier), the PLL is locking on F070RB using the external clock source. This uses the code path on line 92.

therealprof:
Can you please squash the commits?

pigrew:
@therealprof : I've force-pushed a squashed commit. I'll squash things in the future for this repository.

After merging the `SysTick` patch, it looks like CI is passing on this. (Thanks!)

PR #85, opened 2020-01-14
RCC: USB source switching code.

### LINKED ISSUE #83: USB: STM32F070: USB Clock source selection
On the STM32F070RB, USB can only be used with the PLL source, not HSI48 (which doesn't exist). A method to select `RCC.CFGR3.USBSW=1` should be provided.

Some implementation points to be considered (with my tentative answers in parenthesis):

1. What should the default be? (always set to HSI48?)
2. Should there be a debug assertion in `freeze()` that the PLL is enabled if the PLL source is selected for USB? (Yes)
3. Should there be a debug assertion in the USB HAL `enable()` that the selected clock's speed is 48 MHz? (Maybe not since we'd need to start also adding Rcc to the USB peripheral struct)
3. What should the usage be?

Proposed usage:
```
    let mut rcc = dp.RCC
        .configure()
        .usbclksrc(stm32f0xx_hal::rcc::USBClkSource::HSI48)
        .sysclk(48.mhz())
        .hclk(48.mhz())
        .pclk(48.mhz())
        .freeze(&mut dp.FLASH);
```

ADDENDUM: The STM32F1xx HAL has some logic related assertions about the USB clock, so my plan is to duplicate its API as much as possible.

pigrew:
This adds an RCC API feature to set the USB clock source. This is necessary for STM32F070 and also if a crystal is desired to be used for a USB clock (instead of the RCC).

Similar to yesterday's patch, this has lots of `cfg` attributes to control which features are available for the various MCU models. It looks like we could create a new attribute macro like `cfg_has_usb` to apply the proper device list, but I'm leaning towards the simplicity of what this PRcontains.

This does not yet make USB F070 USB work (yet), as the PLL multipliers are not calculated correctly (related to #78), but it is another step in that direction. 

This should close #83.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Added "bypass" parameter to Rcc HSE configuration (breaking change)
+- Add "usbsrc" function to Rcc configuration, used for selecting USB clock source
 
 ### Fixed
 - RCC: Correct code to enable PLL.
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -25,6 +25,15 @@ impl RccExt for RCC {
                 feature = "stm32f098",
             ))]
             crs: None,
+            #[cfg(any(
+                feature = "stm32f042",
+                feature = "stm32f048",
+                feature = "stm32f072",
+                feature = "stm32f078",
+            ))]
+            usb_src: USBClockSource::HSI48,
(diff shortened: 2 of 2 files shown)


pigrew:
@therealprof

Could you review this PR? I think with this one, and fixing PLL multiplier calculation, USB will be functional on the F070.

pigrew:
@therealprof 

A changelog entry has been added (and I squashed/force-pushed the change). Thanks!

therealprof:
No, thank you!

PR #94, opened 2020-02-04
Seal the DAC trait to prevent UB

jamesmunns:
Before this change, the following user code would compile and run without warnings, and would return an uninitialized `[u8; 16]`, which would be UB:

```rust
struct UB;
use crate::stm32::DAC;
impl Pins<DAC> for UB {
    type Output = [u8; 16];
}

// ...

let mut rcc = dp.RCC.configure().sysclk(8.mhz()).freeze(&mut dp.FLASH);
let gpioa = dp.GPIOA.split(&mut rcc);
let mut uninit = dac(dp.DAC, UB, &mut rcc);
```

This change seals the trait and marks it unsafe to prevent outside malicious use, as well as warn maintainers.

### DIFF
--- a/src/dac.rs
+++ b/src/dac.rs
@@ -87,25 +87,29 @@ pub trait DacPin {
     fn enable(&mut self);
 }
 
-pub trait Pins<DAC> {
-    type Output;
+mod sealed {
+    // This trait is unsafe and sealed because we use MaybeUninit to
+    // return an instance of the type, and should only be used on ZSTs.
+    pub unsafe trait Pins<DAC> {
+        type Output;
+    }
 }
 
-impl Pins<DAC> for PA4<Analog> {
+unsafe impl sealed::Pins<DAC> for PA4<Analog> {
     type Output = C1;
 }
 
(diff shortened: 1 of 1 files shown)


jamesmunns:
Oh, this also removes the deprecation warning around the use of `mem::uninitialized()`.

jamesmunns:
After much better advice from @jonas-schievink, we might want to just get rid of MaybeUninit in general. From chat:

```
jschievink: jamesmunns: why not redesign that trait/fn to not require uninitialized?
jamesmunns: I really don't know how to return an arbitrary associated type
I suppose you could also impl the trait in the macro instead, so you have multiple impls with concrete types instead of one blanket impl
I think the blanket impl strays into GAT territory, but I'm not sure
jschievink:
why not give the assoc type a new trait bound like : Channels that is sealed and has a new method or something like that?
unsafe method, that is
```

therealprof:
@jamesmunns Any followup planned?

jamesmunns:
None yet! I'll re add this to my to-do list!

PR #96, opened 2020-02-05
Fix pllsrc bits for HSE

Disasm:
cc @pigrew


Disasm:
Relevant code was introduced in https://github.com/stm32-rs/stm32f0xx-hal/pull/87

pigrew:
Looks correct to me, thanks for noticing.

RCC is getting closer to correct. :)

PR #113, opened 2020-05-12
Reduce the scope of some critical sections in the examples

torkeldanielsson:


Cargo.toml
```diff
@@ -35,7 +35,7 @@ cast = { version = "0.2", default-features = false }
 cortex-m = "0.6"
 cortex-m-rt = "0.6"
 embedded-hal = { version = "0.2", features = ["unproven"] }
-stm32f0 = "0.10"
+stm32f0 = "0.11"
```
therealprof, line 38: Please do this in a separate PR, we also need to add a CHANGELOG entry for it since that is a breaking change.

torkeldanielsson: Ok!

therealprof:
bors try

PR #115, opened 2020-05-20
Make cortex-m-rt a dev-dependency only

Pagten:
None of the library code seems to depend on `cortex-m-rt`, only the examples do.


therealprof:
Good catch, thanks.

Issue #118, opened 2020-08-18
Have PA9/10/11/12 USB pins automatically remap?


brainstorm:
I was going a bit nuts because my board was not USB-enumerating (for a seemingly simple RTIC example) and it turned out to be that I didn't uncomment those:

```rust
        /* Uncomment the following lines if you have a chip in TSSOP20 (STM32F042F)
        or UFQFPN28 (STM32F042G) package
        This code enables clock for SYSCFG and remaps USB pins to PA9 and PA10.
        */
        dp.RCC.apb2enr.modify(|_, w| w.syscfgen().set_bit());
        dp.SYSCFG.cfgr1.modify(|_, w| w.pa11_pa12_rmp().remapped());
```

Fortunately, @mvirkkunen and other very helpful folks at rust-embedded discord channel pointed that out...

Now, would it be at all possible to detect those different variants automatically, somehow?

mvirkkunen:
I think it might be possible by creating a different `Peripheral` struct for the PA9/PA10 pins that's available on certain devices, and having the `enable` function remap them. Or alternatively have a special constructor for it.

therealprof:
I don't see how. The chips are just bonded into different packages and are not self-aware of how they're bonded.

brainstorm:
> ...to do the remapping without having to mess with the registers by hand, that is. It's probably not possible to detect which chips have which pins, because it's package dependent.

Yes, I meant this, sorry for the confusion @therealprof, I should have phrased that better :-S

therealprof:
@brainstorm Thanks for the clarification. Yes, that makes absolute sense.

Want to take a stab at it?

brainstorm:
Sure, I don't know what I'm doing, but let's take it on PR #119 ;)

PR #119, opened 2020-08-18
Move usb-remap internals for certain ICs to USB enable()

brainstorm:
Hopefully fixes issue #118 ?


src/usb.rs
```diff
@@ -28,6 +28,7 @@ unsafe impl UsbPeripheral for Peripheral {
 
     fn enable() {
         let rcc = unsafe { (&*RCC::ptr()) };
```
brainstorm, line 30: Newbie question, @therealprof: can this reference to the pointer of RCC `unsafe` be avoided somehow? I'm curious.

mvirkkunen:
I think it would be preferable to avoid extra feature flags. Would it be possible to implement this using a separate `Peripheral` type and/or constructor?

src/usb.rs
therealprof: Only if you require passing in `RCC` into your function or putting the method on the "owner" of RCC.

therealprof:
I concur with @mvirkkunen and would prefer this to be not dependent on a feature flag.

However the elephant in the room is certainly the enabling of the `SYSCFG` peripheral. It is a quite important peripheral but then again maybe not that important that we would want to unconditionally enable it? Hm...

brainstorm:
Totally broken code passes? hm...

brainstorm:
No worries @therealprof, this got me acquainted with Rust embedded HALs for the first time, so I shall thank you ;)

Tusen tack!

therealprof:
Thanks for sticking around!

brainstorm:
My pleasure! Would you mind pushing a 0.17.1 release for this? :)

therealprof:
Can do.

mvirkkunen:
N.b. the commented out code in the example is missing the arguments

PR #124, opened 2020-11-10
Add I2C slave.

matoushybl:
This adds an I2C slave implementation for I2C1 that I developed for use in a personal project. This is a very much WIP as no support for I2C2 is available and the code is not compliant with anything in embedded-hal.

It'd be nice if this could be discussed here what can be done to make this code mergeable as there are almost none examples of implementing I2C slave for STM32 and other controllers.

### DIFF
--- a/src/i2c_slave.rs
+++ b/src/i2c_slave.rs
@@ -1,0 +1,383 @@
+use super::i2c::{SclPin, SdaPin};
+use super::pac::I2C1;
+
+#[derive(PartialEq)]
+pub enum TransferState {
+    Idle,
+    Addressed,
+    RegisterSet,
+    Receiving,
+    Transmitting,
+}
+
+#[derive(Copy, Clone)]
+pub enum State {
+    DataRequested(u8),
+    DataReceived(u8),
+}
+
+const BUFFER_SIZE: usize = 32;
+
+pub struct I2CSlave<SDA, SCL> {
+    i2c: I2C1,
+    transfer_buffer: [u8; BUFFER_SIZE],
+    transfer_len: usize,
+    buffer_index: usize,
+    register: u8,
+    transfer_state: TransferState,
+    state: Option<State>,
+    _sda: SDA,
+    _scl: SCL,
+}
+
+// direction as specified in the datasheet
+#[derive(PartialEq)]
+pub enum Direction {
+    Write, // slave is receiver
+    Read,  // slave is transmitter
+}
+
+impl From<Direction> for bool {
+    fn from(dir: Direction) -> Self {
+        match dir {
+            Direction::Write => false,
+            Direction::Read => true,
+        }
+    }
+}
+
+impl From<bool> for Direction {
+    fn from(raw: bool) -> Self {
+        if raw {
+            Direction::Read
+        } else {
+            Direction::Write
+        }
+    }
+}
+
+pub enum Status {
+    AddressMatch(Direction),
+    Busy,
+    Timeout,
+    Overrun,
+    ArbitrationLost,
+    BusError,
+    TransferCompleteReload,
+    Stop,
+    NACKReceived,
+    RxNotEmpty,
+    TxDataMustBeWritten,
+    TxEmpty,
+}
+
+impl<SDA, SCL> I2CSlave<SDA, SCL>
+where
+    SDA: SdaPin<I2C1>,
+    SCL: SclPin<I2C1>,
+{
+    pub fn new(i2c: I2C1, address: u8, sda: SDA, scl: SCL) -> Self {
+        let rcc = unsafe { &(*stm32f0xx_hal::pac::RCC::ptr()) };
+        rcc.apb1enr.modify(|_, w| w.i2c1en().enabled());
+        rcc.apb1rstr.modify(|_, w| w.i2c1rst().set_bit());
+        rcc.apb1rstr.modify(|_, w| w.i2c1rst().clear_bit());
+
+        i2c.cr1.write(|w| {
+            w.nostretch()
+                .enabled() // enable clock stretching
+                .anfoff()
+                .enabled() // enable analog filter
+                .dnf()
+                .no_filter() // disable digital filter
+                .errie()
+                .enabled() // error interrupt enabled
+                .stopie()
+                .enabled() // stop interrupt enabled
+                .nackie()
+                .enabled() // nack interrupt enabled
+                .addrie()
+                .enabled() // address match interrupt enabled
+                .rxie() // rx interrupt enabled
+                .enabled()
+                .wupen()
+                .enabled() // wake up when address match
+        });
+
+        // TODO set up timing for nostretch mode
+        // let scll = cmp::max((((48_000_000 >> 1) >> 1) / KiloHertz(100).0) - 1, 255) as u8;
+        // i2c.timingr.write(|w| {
+        //     w.presc()
+        //         .bits(1)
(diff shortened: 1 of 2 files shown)


therealprof (changes requested):
Hi there 👋.

Thanks for this PR. I think with 3 more changes we can merge this:

- [ ] Address the commented out code, including potentially adding some more commentary
- [ ] Add a CHANGELOG entry
- [ ] Add example code so people can directly try the slave functionality

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+        self.transfer_state = TransferState::Idle;
+        self.state = None;
+        self.transfer_len = 0;
+        self.buffer_index = 0;
+        self.i2c.isr.modify(|_, w| w.txe().set_bit()); // flush txdr
+    }
+
+    pub fn set_transmit_buffer(&mut self, buffer: &[u8]) {
+        for (index, item) in buffer.iter().enumerate() {
+            self.transfer_buffer[index] = *item;
```
quarthex, line 405: This may panic since `buffer` can be wider than `self.transfer_buffer`.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
+use crate::i2c::{SclPin, SdaPin};
+use crate::rcc::Rcc;
+use core::ops::Deref;
+
+#[derive(PartialEq)]
```
quarthex, line 5: You should derive from `Eq`, since there is no way for `a == b` being different from `b == a`.
Also make your public types derive from `Debug`, since developers may try to debug their stuff.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+pub enum TransferState {
```
quarthex, line 6: Add a brief documentation for every public type + each of its variant.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+    transfer_buffer: [u8; BUFFER_SIZE],
+    transfer_len: usize,
+    buffer_index: usize,
+    register: u8,
+    transfer_state: TransferState,
+    state: Option<State>,
+    pins: (SCL, SDA),
+}
+
+// direction as specified in the datasheet
```
quarthex, line 33: This comment could be promoted to documentation (`///`).

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+#[derive(PartialEq)]
+pub enum Direction {
+    Write, // slave is receiver
+    Read,  // slave is transmitter
+}
+
+impl From<Direction> for bool {
```
quarthex, line 40: Public conversion from a direction to `bool` does not make much sense. What does a `true` direction mean?
In my opinion, you should use `dir == Direction::Read` for raw bits.
A `Direction::is_read()` would be even cleaner.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+#[allow(dead_code)]
+type I2cRegisterBlock = crate::pac::i2c1::RegisterBlock;
+
+impl<I2C, SCL, SDA> I2CSlave<I2C, SCL, SDA>
+where
+    I2C: Deref<Target = I2cRegisterBlock>,
+{
+    fn i2c_init(self, address: u8) -> Self {
+        self.i2c.cr1.write(|w| {
+            w.nostretch()
```
quarthex, line 137: This looooong call list would be easier to handle if each setting were on its own line.
Something like:
```rs
w.nostretch().enabled(); // enable clock stretching
w.anfoff().enabled(); // enable analog filter
w.dnf().no_filter(); // disable digital filter
...
```
There is no performance issue writing this that way.
This remark applies to following register updates.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+                .bit7()
+        });
+
+        self.i2c.cr1.modify(
+            |_, w| w.pe().enabled(), // enable peripheral
+        );
+        self
+    }
+
+    pub fn is_status(&self, status: Status, clear: bool) -> bool {
```
quarthex, line 187: There are a lot of unnecessary `return`s.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+    feature = "stm32f098",
+))]
+i2c_slave! {
+    I2C2: (i2c2_slave, i2c2en, i2c2rst, apb1enr, apb1rstr),
+}
+// It's s needed for the impls, but rustc doesn't recognize that
+#[allow(dead_code)]
+type I2cRegisterBlock = crate::pac::i2c1::RegisterBlock;
+
+impl<I2C, SCL, SDA> I2CSlave<I2C, SCL, SDA>
```
quarthex, line 131: All public methods should be documented.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+            return;
+        }
+        if self.is_status(Status::NACKReceived, true) {
+            self.handle_error();
+            return;
+        }
+        if self.is_status(Status::Timeout, true) {
+            self.handle_error();
+            return;
+        }
```
quarthex, line 343: This can be merged with a single or-condition.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+            if self.transfer_state == TransferState::Addressed {
+                self.transfer_state = TransferState::RegisterSet;
+                self.register = self.read();
+            } else if self.transfer_state == TransferState::RegisterSet {
+                self.transfer_state = TransferState::Receiving;
+            } else if self.transfer_state == TransferState::Receiving {
+                // do not change state, just read
+                self.transfer_buffer[self.buffer_index] = self.read();
+                self.buffer_index += 1;
+            }
```
quarthex, line 355: A `match` would be more readable.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+                    self.write(self.transfer_buffer[self.buffer_index]);
+                    self.buffer_index += 1;
+                } else {
+                    self.enable_txie(false);
+                    self.state = None;
+                }
+            }
+        }
+        if self.is_status(Status::AddressMatch(Direction::Read), true) {
+            if self.transfer_state == TransferState::RegisterSet {
```
quarthex, line 387: Merge the conditions.

src/i2c_slave.rs
```diff
@@ -0,0 +1,427 @@
...
+            self.i2c.isr.modify(|_, w| w.txe().set_bit()); // flush txdr
+            self.transfer_state = TransferState::Idle;
+        }
+
+        if self.is_status(Status::AddressMatch(Direction::Write), true) {
+            self.transfer_state = TransferState::Addressed;
+        }
+        if self.is_status(Status::TxDataMustBeWritten, false) {
+            // this may be true more times than actual data length, ignore then
+            if self.transfer_state == TransferState::Transmitting {
```
quarthex, line 375: Merge the conditions.

quarthex (changes requested):
This feature would be a great addition to the crate.
Thanks for your work. :+1:

matoushybl:
@quarthex Thanks for the review! This pull request is more than 6 years old, I no longer work with F0s nor do I plan to. If you have a use case for it, you can take over or reimplement it from scratch (given the low code quality). 
For a sample implementation, you can have a look at embassy-stm32 which recently added I2C slave support.

PR #125, opened 2020-11-10
Add CAN bus.

matoushybl:
Adds CAN bus as I developed it for my project.

Any feedback is welcome. I'll continue the development to make it fully working in context of the HAL.

Also CAN traits are under review in embedded-hal, so I might implement the traits before they are merged for future compatibility.


src/i2c_slave.rs
```diff
@@ -0,0 +1,383 @@
...
+        self.state = None;
+        self.buffer_index = 0;
+        self.transfer_len = 0;
+        data
+    }
+
+    pub fn get_state(&self) -> Option<State> {
+        self.state
+    }
+}
```
jonas-schievink, line 383: Unrelated change?

src/lib.rs
```diff
@@ -84,6 +86,12 @@ pub mod usb;
 #[cfg(feature = "device-selected")]
 pub mod watchdog;
 
+// TODO other controllers have CAN
+#[cfg(any(
+    feature = "stm32f091",
```
jonas-schievink, line 91:
```suggestion
    feature = "stm32f091",
    feature = "stm32f042",
    feature = "stm32f072",
```

That's all of them according to the ST website.

matoushybl:
Hi, @jonas-schievink! Thanks for the review! I have fixed the problems you've found and generally fixed some problems with the code. It is still not mergeable to master (pin definitions are still not implemented), but I expect to finish it sometime next week.

therealprof (changes requested):
Missing a CHANGELOG entry. ;)

therealprof (approved):
Haven't had a chance to test it out in person but let's get this in so people can play with it.

Thanks a lot!

bors r+

PR #129, opened 2020-12-06
Update `remap_pins` stuff

david-sawatzke:
I'd also prefer it taking the hal `Rcc`, but that'd be a breaking change. @therealprof Your thoughts?


therealprof (approved):
LGTM

bors r+

brainstorm:
Thanks for fixing this @david-sawatzke, partly my fault ;)

PR #131, opened 2020-12-24
Poll for completion on SPI write.

### LINKED ISSUE #130: SPI blocking transfers do not wait for completion.
# Problem

When doing a SPI transfer with the blocking traits the CS line goes high before the transfer is completed.
```rust
    let mut buf: [u8; 1] = [0];
    w5500_cs.set_low().unwrap();
    spi.write(&[0x00, 0x39, 0x00]).unwrap();
    spi.transfer(&mut buf).unwrap();
    w5500_cs.set_high().unwrap();
```

image

# Expected Behavior

When I add some hacky polling for the `BSY` bit (`0x80`) in the SPI status register to that code, the transfer works as expected, with the CS line going high only after completion of the SPI bus activity.

```rust
    let mut buf: [u8; 1] = [0];
    w5500_cs.set_low().unwrap();
    spi.write(&[0x00, 0x39, 0x00]).unwrap();
    let mut spi1_sr: u32 = unsafe { core::ptr::read_volatile(0x4001_3008 as *const u32) };
    while spi1_sr & 0x80 != 0x00 {
        spi1_sr = unsafe { core::ptr::read_volatile(0x4001_3008 as *const u32) };
    }
    spi.transfer(&mut buf).unwrap();
    let mut spi1_sr: u32 = unsafe { core::ptr::read_volatile(0x4001_3008 as *const u32) };
    while spi1_sr & 0x80 != 0x00 {
        spi1_sr = unsafe { core::ptr::read_volatile(0x4001_3008 as *const u32) };
    }
    w5500_cs.set_high().unwrap();
 ```

image

This seems to be a case of under-specified behavior in the embedded-hal (which is fixed in 1.0.0), the docs do not specify if the `blocking` part is just blocking for buffer space, after which the function returns, or blocking for buffer space, and polling for completion.

I *think* this is a bug though, because despite the under-specification other STM32 embedded HAL traits do block for completion (such as the stm32f1 which I also tested), this appears to be the odd one out.

# Other Information
I am using a STM32F070CBTx.

Relevant `Cargo.toml`:
```
[dependencies]
cortex-m = "0.6.4"
cortex-m-rt = { version = "0.6.13", features = ["device"] }
embedded-hal = "0.2.4"
stm32f0xx-hal = { version = "0.17.1", features = ["stm32f070xb"] }
rtt-target = "0.3.0"
```

```
$ cargo -V             
cargo 1.50.0-nightly (a3c2627fb 2020-12-14)
```

### PR DESCRIPTION
**Pull request by newAM**
Closes #130

---

Continuing discussion from #130 :

The suggested implementation was replacing this:
```rust
 self.check_send().ok(); 
 Ok(()) 
 ```
 
 With this:
 ```rust
 nb::block!(self.check_send())
 ```
 
The concern I have with that is it seems the last `check_send().ok()` was intended to clear the `OVR` error bit that may be set due to dropped reads.

Additionally, it does not seem that this works as intended, the CS still goes high before the write is completed with consecutive write calls.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -400,7 +400,7 @@ where
             nb::Error::Other(Error::ModeFault)
         } else if sr.crcerr().bit_is_set() {
             nb::Error::Other(Error::Crc)
-        } else if sr.txe().bit_is_set() {
+        } else if sr.txe().bit_is_set() && sr.bsy().bit_is_clear() {
             return Ok(());
         } else {
             nb::Error::WouldBlock
@@ -481,15 +481,8 @@ where
             bufcap -= 1;
         }
 
-        loop {
-            let sr = self.spi.sr.read();
-            if !sr.bsy().bit_is_set() {
-                break;
-            }
-        }
-
         // Do one last status register check before continuing
-        self.check_send().ok();
+        nb::block!(self.check_send()).ok();
         Ok(())
     }
 }
@@ -532,15 +525,8 @@ where
             self.send_u16(word.clone());
         }
 
-        loop {
-            let sr = self.spi.sr.read();
-            if !sr.bsy().bit_is_set() {
-                break;
-            }
-        }
-
         // Do one last status register check before continuing
-        self.check_send().ok();
+        nb::block!(self.check_send()).ok();
         Ok(())
     }
 }
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -481,6 +481,13 @@ where
             bufcap -= 1;
         }
 
+        loop {
+            let sr = self.spi.sr.read();
+            if !sr.bsy().bit_is_set() {
+                break;
+            }
+        }
+
         // Do one last status register check before continuing
         self.check_send().ok();
         Ok(())
@@ -523,6 +530,13 @@ where
         for word in words {
             nb::block!(self.check_send())?;
             self.send_u16(word.clone());
+        }
+
+        loop {
+            let sr = self.spi.sr.read();
+            if !sr.bsy().bit_is_set() {
+                break;
+            }
         }
 
         // Do one last status register check before continuing

### CODE REVIEW

**Comment by therealprof** [comment]
For this to work `check_send` might potentially need to be changed to first check for `bsy` and return `nb::Error::WouldBlock` if that's the case. If there's room in the buffer it will not return `nb::Error::WouldBlock` which is what we want so we can block on it.

**Comment by therealprof** [comment]
We can also combine the current current and the `nb::block!` in case we want to retain the current behaviour of ignoring any errors at that point.

**Comment by newAM** [comment]
Upon looking at the code again the final `check_send().ok()` does not appear to be clearing the `OVR` flag.

The STM provided HAL (in C) *does* clear the OVR flag when operating in write-only mode, but the procedure to clear that flag  requires reading DR, then reading SR.

> When an overrun condition occurs, the newly received value does not overwrite the previous one in the RXFIFO. The newly received value is discarded and all data transmitted subsequently is lost. Clearing the OVR bit is done by a read access to the SPI_DR register followed by a read access to the SPI_SR register.

---

> For this to work `check_send` might potentially need to be changed to first check for `bsy` and return `nb::Error::WouldBlock` if that's the case. If there's room in the buffer it will not return `nb::Error::WouldBlock` which is what we want so we can block on it.

Along the lines of this?

```rust
    fn check_send(&mut self) -> nb::Result<(), Error> {
        let sr = self.spi.sr.read();

        Err(if sr.ovr().bit_is_set() {
            nb::Error::Other(Error::Overrun)
        } else if sr.modf().bit_is_set() {
            nb::Error::Other(Error::ModeFault)
        } else if sr.crcerr().bit_is_set() {
            nb::Error::Other(Error::Crc)
        } else if sr.txe().bit_is_set() && sr.bsy().bit_is_clear() {
            return Ok(());
        } else {
            nb::Error::WouldBlock
        })
    }
```

therealprof:
> Along the lines of this?

Yes, something like that might work. I'm actually not sure which other conditions would be reason to use `nb::Error::WouldBlock`. Maybe we should just explictly for the `bsy` condition up front and return `nb::Error::WouldBlock` and turn the `else` into a hard Error.

therealprof:
@newAM Ping?

newAM:
Sorry for distracted with other projects :sweat_smile: 

I made the modification - both the first and second implementations fix the issues mentioned (`BIDIMODE` / `BIDIOE` race condition & back to back calls).

> Yes, something like that might work. I'm actually not sure which other conditions would be reason to use nb::Error::WouldBlock. Maybe we should just explictly for the bsy condition up front and return nb::Error::WouldBlock and turn the else into a hard Error.

I am not sure of the other conditions either, which makes me hesitant to change the existing implementation too much.  It does feel like this could be optimized better, but I have not spent enough time with this hardware to be confident about that.

src/spi.rs
```diff
@@ -482,7 +482,7 @@ where
         }
 
         // Do one last status register check before continuing
-        self.check_send().ok();
+        nb::block!(self.check_send()).ok();
```
WasabiFan, line 485: Still `.ok()` rather than returning the result?

therealprof (approved):
LGTM, thanks for bearing with me.

bors r+

newAM:
> LGTM, thanks for bearing with me.

Thanks for bearing with me as well :+1: 

Happy to help!

src/spi.rs
newAM: I left it as `ok()` because that better preserves the previously existing behavior.

...as for whether or not it _should_ be `ok()` or `?` that is debatable.  What do you think?

I _think_ the previous intent was to clear the overrun flag if it was set, but that requires a read to DR followed by a read access to SR to clear.

WasabiFan: My understanding is that before it had _no_ tangible effect (since it didn't read from DR) and now it blocks but still does not affect the flags. It seems a bit strange if we were to call this function and identify a failure but not report it -- then again, if it didn't prevent transmission, perhaps it's best to ignore it. It may just depend on the error.

All in all, I don't really have an influencing perspective here since I'll be cherry-picking and tacking on my own changes in my project's fork (I need the non-blocking behavior in some cases). The `.ok()` pattern to ignore an error seems strange to me since it looks more like a mistake, but if this is the behavior that works, I can't say much against it!

Thanks for fixing this! 😄

PR #137, opened 2021-04-22
Update bxcan to 0.5.0

marcelbuesing:
[bxcan 0.5.0](https://github.com/stm32-rs/bxcan/blob/master/CHANGELOG.md) updates the defmt dependency to 0.2.0.


therealprof (commented):
Would you mind adding a CHANGLOG entry for a breaking change?

thejpster:
0.6.0 is out now.

PR #138, opened 2021-05-24
PWM support

### LINKED ISSUE #26: PWM support

jonfin:
Add PWM support based on the stm32f4xx_hal implementation. It is mostly a copy from stm32f4xx_hal with some modification.
Only change to stm32f4xx_hal: it uses the recommended order for setting the register from DM00042534

I did some testing based on a stm32f030c8 chip.

fixes #26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Add CAN bus abstraction based on the [bxcan] crate.
+- Add PWM output generation based on the timers.
 
 [bxcan]: https://crates.io/crates/bxcan
 
--- a/examples/pwm.rs
+++ b/examples/pwm.rs
@@ -1,0 +1,38 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+// Halt on panic
+use panic_halt as _;
+
+use cortex_m_rt::entry;
+
+use stm32f0xx_hal as hal;
+
+use hal::{pac, prelude::*, pwm};
+
+#[entry]
+fn main() -> ! {
+    if let Some(mut dp) = pac::Peripherals::take() {
+        // Set up the system clock.
+        let mut rcc = dp.RCC.configure().sysclk(8.mhz()).freeze(&mut dp.FLASH);
+
+        let gpioa = dp.GPIOA.split(&mut rcc);
+        let channels = cortex_m::interrupt::free(move |cs| {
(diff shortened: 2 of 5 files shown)


therealprof:
Thanks for the PR! I've tested it on a STM32F072 and it works a treat. If you could fix the CI issues and register the example with `Cargo.toml` I can merge this in.

jonfin:
It should now work even on more devices. I missed the possibility that there could be other devices than the stm32f030 or stm32f070.

One thing is missing: the PWM implementation for TIM2 (32-bit). I skip it because I cannot test it after changing the calculation of the prescaler / Auto-reload value.

therealprof:
Thanks, it's still missing the example registration in `Cargo.toml`. And if you could squash your PR a bit that'd be much appreciated.

jonfin:
Is the example not automatic registered?

therealprof:
Ah sorry, I missed that it's an example that works on all chips; not used to that. 😅

jonfin:
Me neither ;)
I squashed the commits.

therealprof (approved):
LGTM, thanks.

bors r+

PR #143, opened 2021-10-31
Update stm32f0 peripheral access crate from 0.13 to 0.14

newAM:
The clippy failures look like a pre-existing condition.  I can clean those up in a separate PR if desired.


therealprof (approved):
LGTM, thanks!

PR #150, opened 2022-02-13
Set mode correctly on C2 of 2-channel PWM

Gekkio:
When configuring a 2-channel PWM and setting up channel 2 mode, we accidentally set up channel 1 mode instead.

I haven't actually verified the currently broken behaviour on hardware since I'm using 4-channel PWM where this configuration is done correctly, but this one looks like a simple copy-paste error which I spotted while studying the code :sweat_smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


newAM (approved):
Looks good to me.

CI is failing for unrelated reasons, opened a fix here: #152

therealprof:
bors try

therealprof:
bors r=newAM

PR #159, opened 2023-01-23
Provide access to serial status flags idle/txe/rxne

oll3:
Noticed that the serial interface were missing access to these flags and seem to be present in other stm32xx hals (like stm32f4xx-hal).

Added them so that an ISR can differ between these events.

Edit: removed unnecessary unsafe.


therealprof:
Mind adding a CHANGELOG entry please?

oll3:
> Mind adding a CHANGELOG entry please?

Sure. Added a getter for the tc (transmission complete) flag too.

therealprof (approved):
LGTM, thanks.

bors r+

PR #161, opened 2023-04-22
Provide ability to clear timer UIF interrup flag

revyakin:
Timer interrupts cannot be used without this method.


therealprof (approved):
LGTM, thanks.

bors r+

PR #164, opened 2023-07-18
Copy NoSck, NoMiso and NoMosi implementations from stm32f1xx-hal

JanekGraff:
This allows using SPI without all pins needing to be defined.


JanekGraff:
Sry for making this, this was a missclick of something i was testing that i wanted to merge into my master.

PR #169, opened 2023-08-19
New Interrupt Driven Serial Echo Example

AdinAck:
## Motivation
- The provided `serial_echo` example seemed lacking imho
- Would be really nice if it used interrupts

## Change(s)
- Added `serial_echo_irq.rs` to `/examples`


AdinAck:
This compiles and runs on my end... looking into it.

AdinAck:
that first error makes no sense, still compiles in my deployment repo but not in this repo... 🤷‍♂️

therealprof:
Adding an example seems fine to me, though this is unnecessarily complex and convoluted:

- The macros (just to deal with the global variables) seem a bit over the top and make it even harder to follow for the uninitiated.
- Since you're toggling the LED in the main loop, there's no need for all that extra global complexity, a local variable will do.

PS: If you want to use a resource only in a single place (like an interrupt handler), you can also move it into a variable in the interrupt handler once and then safely use  it without a critical section.

AdinAck:
> * The macros (just to deal with the global variables) seem a bit over the top and make it even harder to follow for the uninitiated.

Yes those were left over from my main project where I have many global variables and want it to be easy to use them later, I will remove this.

> * Since you're toggling the LED in the main loop, there's no need for all that extra global complexity, a local variable will do.

Haha another remnant from my main project, I will fix this as well.

> PS: If you want to use a resource only in a single place (like an interrupt handler), you can also move it into a variable in the interrupt handler once and then safely use it without a critical section.

I have no idea what this means, and if you are able to explain to me further at your convenience, I would greatly appreciate it.

examples/serial_echo_irq.rs
```diff
@@ -0,0 +1,102 @@
...
+                /*
+                Failed to read a byte:
+
+                There could be some kind of alignment error or the UART
+                was disconnected or something.
+                 */
+            }
+        }
+    });
+}
```
therealprof, line 102: What I mean is that this is really frowned upon. `cortex_m::interrupt::free` disables interrupts within the interrupt handler and with interrupts disabled you additionally execute a slow operation...

If you only use a resource in an interrupt handler, you can move it into the handler and only use the `cortex_m::interrupt::free` once for initial purposes and have the remaining executions of the interrupt handler run lock free. cf. examples/blinky_timer_irq.rs

AdinAck: Aaaahh, yes wonderful. I love that. I have been wondering what the proper "pattern" is for this, I will update the example to reflect the proper way of doing this.

AdinAck:
I believe I have correctly converted the example over to the proper way of using the global serial port. I have also rebased this PR.

AdinAck:
#170 will fix this PR.

PR #170, opened 2023-11-24
Fix compilation error in spi.rs for rust Version 1.73+

JanekGraff:
With rust version 1.73.0 the `invalid_reference_casting` lint was set to deny by default. (See [PR#112431](https://github.com/rust-lang/rust/pull/112431). This causes the build process of this crate to fail to build spi.rs (only when building the crate with its examples, when adding it as a dependency works fine from what i can see).

This change fixes the compilation by adding the `[allow(invalid_reference_casting)]` attribute to the relevant parts.

This also fixes a `unused doc comment` compilation warning in rcc.rs

# NOTE

I have not tested this on hardware, just added the attribute to fix the compilation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Wrong mode when using PWM channel 2 of a two-channel timer
 - `adc_values` example conversion error
+- `invalid_reference_casting` Compilation error in spi.rs for Rust version 1.73+ (
+  See [PR#112431](https://github.com/rust-lang/rust/pull/112431) for more info)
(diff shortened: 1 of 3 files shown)


adamgreig:
See also https://github.com/stm32-rs/stm32f7xx-hal/pull/207 and https://github.com/stm32-rs/stm32f3xx-hal/pull/346 which had the same problem and a couple other options for fixing it.

JanekGraff:
> @JanekGraff Thanks for trying to fix this. Would you mind changing to `core::ptr::addr_of!` instead of squelching the warning?
> 
> Unrelated nit: Could you please remove the added newlines between the links in the CHANGELOG.md?

Done

therealprof (approved):
LGTM, thanks!

PR #172, opened 2023-12-08
Implement interface to internal flash storage

JanekGraff:
This implements an interface to reading and writing to the internal flash memory. This implements the [`NorFlash` traits](https://github.com/rust-embedded-community/embedded-storage/blob/master/src/nor_flash.rs) from the [`embedded-storage` crate](https://docs.rs/embedded-storage/latest/embedded_storage/).

The interface is inspired by the interfaces used in the [`stm32f4xx-hal`](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/flash.rs) and the [`stm32g0xx-hal`](https://github.com/stm32-rs/stm32g0xx-hal/blob/main/src/flash/mod.rs). 

The interface could definitely use some clean up and I'm open to suggestions on that matter. Especially the handling of [writing to addresses not aligned to the native write size](https://github.com/JanekGraff/stm32f0xx-hal/blob/2769e82dfa8e42b74f7e367601d5bc5d2873d36d/src/flash.rs#L231) is something I'm not sure about. The `stm32g0xx-hal` has a [different approach](https://github.com/stm32-rs/stm32g0xx-hal/blob/e6145ef83c639e736556fb4775f349aef7100ad6/src/flash/mod.rs#L166) for this, but their approach has some problems in my opinion:

1. Trying to write data that is shorter than the unaligned size (e.g. writing 3 Bytes to `0x08008001`) will lead to a panic due to the slice index being out of range.
2. *(I'm not 100% sure about this)* It seems like they will write unaligned data to an address that is in front of the address given by the user, which is not intuitive in my opinion.
3. Passing an unaligned address to the write function will lead to some of the data in front of the address being overwritten.

The reasons above are why i choose the approach of returning an Error instead of trying to pad the data into native writes.

I also provided a [basic example](https://github.com/JanekGraff/stm32f0xx-hal/blob/flash/examples/flash_read_write.rs) of how to use the flash interface.

The CI checks of this PR will fail until #170 (or an alternative solution) is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Provide getters to serial status flags idle/txe/rxne/tc.
 - Provide ability to reset timer UIF interrupt flag
 - PWM complementary output capability for TIM1 with new example to demonstrate
+- Implement interface for reading and writing to the internal flash memory and an example for demonstration.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ nb = "1"
 void = { version = "1.0", default-features = false }
 stm32-usbd = { version = "0.6", optional = true }
 bxcan = "0.6.0"
+embedded-storage = "0.3.0"
 
 [dev-dependencies]
 cortex-m-rt = "0.7"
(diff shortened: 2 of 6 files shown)


therealprof (approved):
Very nice!

PR #176, opened 2024-03-26
impl embedded_hal_1::i2c::I2c for I2c

dimpolo:


src/i2c.rs
```diff
@@ -456,3 +435,93 @@ where
...
+    ) -> Result<(), Self::Error> {
+        self.write_read_impl(address, write, read)
+    }
+
+    fn transaction(
+        &mut self,
+        _address: embedded_hal_1::i2c::SevenBitAddress,
+        _operations: &mut [embedded_hal_1::i2c::Operation<'_>],
+    ) -> Result<(), Self::Error> {
+        todo!()
```
TeXitoi, line 506: We can’t let that as is. It must be implemented.

TeXitoi (changes requested):

PR #180, opened 2024-08-22
Use `critical_section::CriticalSection` for GPIO configuration

urdh:
This change improves interoperability with RTICv2 (which exposes a `bare_metal::CriticalSection` during initialization).


TeXitoi:
We can’t do that as is. We need at least to provide a safe way of getting GPIOs.

Do you have any reference on another hal doing something similar, or a guideline about the removal of the token in `cortex_m::interrupt::free`?

urdh:
As far as I can tell, most HALs don't require a critical section at all when reconfiguring GPIOs.

Upstream actually [recommends `critical_section::with`](https://github.com/rust-embedded/cortex-m/blob/master/cortex-m/CHANGELOG.md#unreleased) as a replacement for `cortex_m::interrupt::free`, but that provides a different `CriticalSection` which is incompatible with the `bare_metal` one. Still, this crate should probably use `critical_section` rather than `bare_metal`.

(I guess ultimately I should log an issue with RTIC ([again](https://github.com/rtic-rs/rfcs/issues/32)) to have RTIC use *critical-section*; in the meantime interfacing between the two will require a tiny bit of `unsafe` code.)

urdh:
Ping! This now uses `critical_section` as recommended by upstream as per https://github.com/stm32-rs/stm32f0xx-hal/pull/180#issuecomment-2515389384 above (though the correct reference is [on the *c-m-in* branch](https://github.com/rust-embedded/cortex-m/blob/c-m-in/cortex-m/CHANGELOG.md#unreleased)), and RTIC has moved to `critical_section` in https://github.com/rtic-rs/rtic/pull/1120 so there shouldn't be any incompatibility once their next release drops.

PR #189, opened 2025-11-26
gpioe: AF1 enablement

jakobjung10:
+ This commit reenables AF1 for pin 13, 14, 15 on GPIOE which is necessary for SPI communication.


therealprof (approved):
LGTM, thanks.

PR #194, opened 2026-01-24
Fix incorrect OSPEEDR register assignment for `into_push_pull_output_hs`

### LINKED ISSUE #191: GPIOx_OSPEEDR is set incorrectly
See the following code:
https://github.com/stm32-rs/stm32f0xx-hal/blob/8eda8641428e901c4987aca292a9c4c81f9ca63c/src/gpio.rs#L487-L489

stm32f0xx-hal assumes 1 bit per GPIO, but according to the "STM32F030x4/x6/x8/xC and STM32F070x6/xB" datasheet, this register has 2 bits per GPIO:

Image

So the linked line should be like so:
```rs
w.bits(r.bits() | (0b11 << offset))
```

holly-hacker:
Fixes #191

This code appears to be incorrectly copied from the OTYPER assignment above it.


therealprof (approved):
LGTM, thanks.

